PR #1, opened 2021-05-13
UTF-8 Encoded Text

emilypi:
This proposal outlines a project plan for the migration of the `text` package from its current default encoding (UTF-16) to a new default of UTF-8.  

The lack of UTF-8 as a default in the `text` package is a pain point raised by the Haskell Community and
many of our industry partners for many years. We have done our homework in soliciting feedback from the broader community and industry, and have received positive affirmation of the following: 

1. **This change is desirable:** Overwhelmingly, the community was positive on such a change, and industry was the same. 
2. **There is an appetite for breakage:** neither the community nor the people we spoke to in industry were concerned about the cost of migration or breakage, though, as a deliverable, we will make sure to minimize the cost of migration by reusing the `text` package name, and providing alternatives in the case that users require UTF-16 text. 
3. **The text maintainers are on board:** All text maintainers currently support this effort, and will be reviewers and support for the project leader throughout the course of the project. 

[Rendered](https://github.com/haskellfoundation/tech-proposals/blob/bd0828f515e3d7746061f218f7b7856a912fe8a1/proposals/002-text-utf-default.md)

### DIFF
--- a/proposals/002-text-utf-default.md
+++ b/proposals/002-text-utf-default.md
@@ -86,8 +86,8 @@ The very `instance Binary Text` serializes `Text` in UTF-8 encoding.
 If we switch the internal representation of `Text` from UTF-16 to UTF-8,
 all such conversions would be made redundant and we'll be able just check that
 a `ByteString` is a valid UTF-8 (which is most often the case) and copy it into `Text`.
-If in future (see an upcoming "Unifying vector-like types" proposal) `ByteString`
-is backed by unpinned memory, we'd be able to eliminate copying entirely.
+If in future `ByteString` switch to be
+backed by unpinned memory, we'd be able to eliminate copying entirely.
 
 `Text` is also often used in contexts, which involve mostly ASCII characters.
 This often prompts developers to use `ByteString` instead of `Text` to save 2x space
@@ -117,6 +117,8 @@ to make another attempt.
 -   Ensure stakeholders (e.g. GHC, Cabal, Stack, boot libs) have ample time to migrate and address any bugs.
 
 -   Performance satisfies targets listed below in "Performance impact" section.
+
+-   Compatibiltiy story satisfies targets listed below in "Compatibility issues" section.
 
 # People
 
@@ -179,6 +181,70 @@ possible. This candidate should be shared publicly and loudly.
 **Implementation:**
 
 -   TBD: There is a straightforward implementation, but this one is left up to Andrew for comment.
+
+**Compatibility issues**
+
+`text` is a very old package, deeply ingrained in Haskell ecosystem.
+A change of internal representation is necessarily a breaking change.
+Our strategy to tackle compatibility issues is guided by a desire
+to finish this project in a time-bound fashion with realistic expectations
+about available resources.
+
+Current `text` HEAD supports GHCs back to GHC 8.0.
+At the moment we do not foresee any blockers
+to keep compatibility with GHC 8.0 after UTF-8 transition,
+and we plan to stick to it even if it causes some overhead in CPP.
+However, if we discover that supporting old GHCs causes a significant overhead
+(e. g., a dedicated non-trivial code path, emulating missing primop
+or working around a bug), we may decide to shrink the compatibility window.
+Such decision would not to be taken lightly, but we believe that getting
+things done for the bright future should not be hindered by old unsupported luggage.
+
+One suggestion to improve compatibility story was to keep both UTF-16 and UTF-8
+implementations in `text` and switch between them via Cabal flag. It seems,
+however, that such strategy will put an undue, indefinitely long burden
+on `text` maintainers, and brings little benefits to downstream packages, because
+they cannot detect build flags of `text` (and thus cannot rely on its internals at all).
+
+Instead we mark a new, UTF-8 release as `text-2.0`, and put a call for volunteers
+to maintain a legacy UTF-16 package. Depending on a demand, this could be done either
+as a continuation of `text-1.X` series, or as a separate `text-utf16` package. We'll facilitate such community project and will work with Hackage Trustees and Stackage
+Curators to ensure timely transition of ecosystem.
+
+With regards to API compatibility, we intend to keep signatures of non-`Internal`
+modules unchanged, except `Word16` replaced by `Word8` where appropriate.
+Such promise unfortunately cannot be made for `Internal` modules,
+due to their nature: even while we'll strive to keep as much untouched as possible,
+the semantics of internal functions is due to change drastically. This kind of breakage
+should not come as a big surprise, because `Internal` modules have a disclaimer about
+unstable API.
+
+There are two places where `text` leaks details of internal represenation.
+First of them is `Data.Text.Array`, which provides an access to an underlying bytearray.
+Not only its API is to change from `Word16` to `Word8`, but also the semantics
+of array switches from UTF-16 to UTF-8. This will cause breakage of several packages
+such as `unicode-transforms` and `unicode-collation`. We intend to communicate with
+respective maintainers as early as possible to help with transition.
+
+Another one is `Data.Text.Foreign`, which is mostly used by `text-icu` library,
+which binds to `libicu` for certain Unicode manipulations. `libicu` provides
+helpers to convert C strings
+[from UTF8 to UTF16](https://unicode-org.github.io/icu/userguide/strings/utf-8.html).
+It is up to `text-icu` maintainers either to modify their bindings. We intend
+to reach to them as soon as we have an MVP.
+
+Since fixing downstream compatibility issues is up to external counterparties,
+most of which are unpaid volunteers, we cannot expect them to do it in a limited
+time frame. We are devoted to having a smooth migration story and will provide
+as much guidance as possible, but to keep our targets time-bound we cannot tie the success
+of this project to actions of third parties. We will not wait for everyone to migrate.
+
+To sum up, we plan to:
+
+* Keep `text` compatible with GHCs back to 8.0, unless it puts an undue cost (more than 50 lines of code per major release).
+* Keep signatures of non-`Internal` modules compatible modulo `Word16`/`Word8` change.
+* Provide migration guidance to clients of `Data.Text.{Array,Foreign}`.
+* Facilitate a community project to keep UTF16-based legacy fork alive, if there is such demand.
 
 **Performance impact**
 
@@ -265,20 +331,20 @@ packages that go out of date.
 
 -   text-2.0.0.0, which will provide a UTF-8 encoding for Text as a default for all versions going forward.
 
--   A `text-utf16` package, which is a preservation of the current UTF-16 encoded text, for backwards compatibility.
-
--   Updates to the Text Haddocks that reflect the UTF-8 changes
+-   Updates to the Text Haddocks that reflect the UTF-8 changes.
 
 -   Announcements and updates across all Haskell channels covering the following:
-    -   Significant dates and milestones
+    -   Significant dates and milestones.
 
-    -   Expected code impact
+    -   Release candidates.
 
-    -   Release candidates
+    -   Migration guides.
 
-    -   Delivery
+    -   Performance impact analysis.
 
--   Migration instructions and design documentation
+    -   Delivery.
+
+    -   Design documentation.
 
 # Outcomes
 
--- a/proposals/002-text-utf-default.md
+++ b/proposals/002-text-utf-default.md
@@ -1,0 +1,156 @@
+This proposal is for the migration of the `text` package from its
+default UTF-16 encoding to UTF-8. The lack of UTF-8 as a default in the
+`text` package is a pain point raised by the Haskell Community and
+industry partners for many years. The Haskell Foundation is uniquely
+positioned to effect a change like this, granted there is an appetite
+for breakage, and that such a project is well-socialized and planned.
+During the meetings of the Haskell Foundation Tech Agenda Track, we
+identified the pros and cons of such a migration against historical
+attempts, and agreed that it was within our power to do.
+
+Further, we solicited both community and industry feedback to gauge the
+appetite for breakage, and what stakeholders would be affected by the
+changes. Andrew Lelechenko has offered to lead the implementation of
+this project.
+
+# Motivation
+
+-   UTF-16 by default requires that all Text values pay a premium for serialization. Arguably, the performance impact of Text is flipped
+    upside-down: most text is UTF-8, and Haskell devs pay an undue cost when working with the wrong default.
+
+-   UTF-8 is the industry standard and by far the most common text encoding, with roughly 97% of web pages existing in UTF-8. The
+    existing UTF-16 default imposes an additional hurdle to working with the vast majority of web content on earth.
+
+-   Many systems in Haskell are UTF-8 by default (e.g. Haddock)
+
+# Goals
+
+-   Solicit feedback from community members and industry to gauge appetite for such a change.
+
+-   Provide an implementation, migration, and delivery plan for changing the default encoding of `text` from UTF-16 to UTF-8.
+
+-   Ensure stakeholders (e.g. GHC, Cabal, Stack, boot libs) have ample time to migrate and address any bugs.
+
+-   Implementation should not significantly alter the performance characteristics of the base `text` library within some tolerance
+    threshold.
+
+# People
+
+-   Performers:
+
+    -   Leader: Andrew Lelechenko (bodigrim)
+
+    -   Support: Emily Pillmore (emilypi)
+
+-   Reviewers:
+
+    -   The text maintainers
+
+        -   Xia Li-Yao (lysxia)
+
+        -   Emily Pillmore (emilypi)
+
+        -   Dan Cartwright (chessai)
+
+        -   Callan McGill (boarders)
+
+-   Stakeholders:
+
+    -   Edward Kmett: has been vocal about his use of `text-icu` and requires it not be broken.
+
+-   Ben Gamari: integration with GHC
+
+-   The Cabal maintainers (fgaz, emilypi, mikolaj): integration with Cabal
+
+Progress will be reported on a weekly basis to the HF Technical Agenda
+Track, with Emily as support for Andrew.
+
+# Timeline
+
+We expect that this project will take roughly 6 months to fully
+complete: 3-4 months to complete the code implementation, performance
+testing, and unit testing, another 1-2 months to integrate with
+stakeholders and diagnose any potential issues with the migration.
+
+**Preparation:**
+
+Using the HVR's existing [`text-utf8`](https://github.com/text-utf8) as
+a starting point, the following must be done before an implementation is
+started:
+
+-   Modernize the codebase and clear out the bitrot
+
+-   Establish a baseline for performance and any related issues.
+
+-   Update testing and performance benchmarks to make use of `inspection-testing` to ensure fusion is not broken in
+    subsequent UTF-8 related changes.
+
+An MVP should completely preserve standard user-facing API, and not
+break fusion. Performance should not significantly diverge from the
+existing UTF-16 text package. There will be an expected change to the
+exposed Text internals, in which case, breakage should be assessed by
+circulating a git commit reference to a release candidate as soon as
+possible. This candidate should be sourced publicly and loudly.
+
+**Implementation:**
+
+-   TBD: There is a straightforward implementation, but this one is left up to Andrew for comment.
+
+**Stakeholders:**
+
+-   Library authors will need to be made aware of changes and adjust accordingly. HF will provide a git reference to a complete MVP as
+    soon as possible, and produce a migration guide.
+
+-   In the case of GHC, Cabal, and other core infrastructure, we will work closely with these packages to help migrate and assess+fix
+    breakages.
+
+While we do not expect many authors to experience significant changes,
+there will be some help that needs to be given in terms of bumping
+Hackage bounds since this migration will be a major version bump. HF
+will need to coordinate with the Hackage Trustees to help move along
+packages that go out of date.
+
+# Deliverables
+
+-   text-2.0.0.0, which will provide a UTF-8 encoding for Text as a default for all versions going forward.
+
+-   A `text-utf16` package, which is a preservation of the current UTF-16 encoded text, for backwards compatibility.
+
+-   Updates to the Text Haddocks that reflect the UTF-8 changes
+
+-   Announcements and updates across all Haskell channels covering the following:
+    -   Significant dates and milestones
+
+    -   Expected code impact
+
+    -   Release candidates
+
+    -   Delivery
+
+-   Migration instructions and design documentation
+
+# Outcomes
+
+-   Addresses a recognized need and want from both Industry and the Haskell Community
+
+-   Better UX for Haskell's text story
+
+-   Establishes the ability for Haskell Foundation to Get Things Done that were previously blocked for 10+ years in the community.
+
+-   Better interoperability and web story
+
+# Risks
+
+-   HF must minimize the cost to migrate, or people will just get mad (and rightfully so).
+
+-   Text-icu will need a bespoke UTF-8 conversion function. In general, the Unicode story must be tracked and made sure it will not break.
+    -   Recommendation: make this a high-priority deliverable when project planning
+
+-   The old UTF-16 text package will need to be preserved, and will require a maintainer.
+
+-   Performance expectations should be managed: UTF-8 text is \*not\* a panacea. It will not result in a 2x or even significant
+    performance increase in many cases. In fact, performance may regress in some programs.
+    -   Recommendation: we must set expectations early and often throughout the implementation process. We expect there to be
+        improvements to performance \*on the margins\*.
+
+    -   Note: We have made this argument, and the appetite for change did not shift, but it is still important to track.
--- a/proposals/002-text-utf-default.md
+++ b/proposals/002-text-utf-default.md
@@ -144,9 +144,9 @@ to make another attempt.
 
     -   Edward Kmett: has been vocal about his use of `text-icu` and requires it not be broken.
 
--   Ben Gamari: integration with GHC
+    -   Ben Gamari: integration with GHC
 
--   The Cabal maintainers (fgaz, emilypi, mikolaj): integration with Cabal
+    -   The Cabal maintainers (fgaz, emilypi, mikolaj): integration with Cabal
 
 Progress will be reported on a weekly basis to the HF Technical Agenda
 Track, with Emily as support for Andrew.
@@ -284,7 +284,7 @@ of avoiding conversion between UTF-8 and UTF-16 conversion.
 Fast validation of UTF-8 is not a trivial task, but we intend to employ
 [`simdjson::validate_utf8`](https://arxiv.org/pdf/2010.03090.pdf) for this task.
 
-Another important aspect of `text` performance is fusion.We are finalising
+Another important aspect of `text` performance is fusion. We are finalising
 an `inspection-testing`-based [test suite](https://github.com/haskell/text/pull/337) to check that
 pipelines, which used to fuse before, are fusing after UTF-8 transition as well.
 Fusion is incredibly fragile matter: for example, of 100 tests, which fuse in GHC 8.10.4,
@@ -329,9 +329,9 @@ packages that go out of date.
 
 # Deliverables
 
--   text-2.0.0.0, which will provide a UTF-8 encoding for Text as a default for all versions going forward.
+-   `text-2.0.0.0`, which will provide a UTF-8 encoding for Text as a default for all versions going forward.
 
--   Updates to the Text Haddocks that reflect the UTF-8 changes.
+-   Updates to the `text` Haddocks that reflect the UTF-8 changes.
 
 -   Announcements and updates across all Haskell channels covering the following:
     -   Significant dates and milestones.
@@ -360,7 +360,7 @@ packages that go out of date.
 
 -   HF must minimize the cost to migrate, or people will just get mad (and rightfully so).
 
--   Text-icu will need a bespoke UTF-8 conversion function. In general, the Unicode story must be tracked and made sure it will not break.
+-   `text-icu` will need a bespoke UTF-8 conversion function. In general, the Unicode story must be tracked and made sure it will not break.
     -   Recommendation: make this a high-priority deliverable when project planning
 
 -   The old UTF-16 text package will need to be preserved, and will require a maintainer.
--- a/proposals/002-text-utf-default.md
+++ b/proposals/002-text-utf-default.md
@@ -57,7 +57,7 @@ no longer know beforehand, where *n*-th code point starts. One must *parse* all 
 from the very beginning to learn which ones are 2-byte long and which are 4-byte long.
 
 But once we abandon requirement of constant indexing, even better option arises. Let's
-encode first 128 characters as 1 byte, some others as 2 bytes, and the rest as 4 bytes.
+encode first 128 characters as 1 byte, some others as 2 bytes, and the rest as 3 or 4 bytes.
 This is UTF-8. The killer feature of this encoding is that it's fully backwards compatible
 with ASCII. This meant that all existing ASCII documents were automatically valid UTF-8
 documents as well, and that 50-years-old executables could often parse UTF-8 data without
@@ -90,11 +90,19 @@ If in future (see an upcoming "Unifying vector-like types" proposal) `ByteString
 is backed by unpinned memory, we'd be able to eliminate copying entirely.
 
 `Text` is also often used in contexts, which involve mostly ASCII characters.
-For such applications storing data in UTF-8 means using up to 2x less space,
-which could be important to reduce memory pressure.
+This often prompts developers to use `ByteString` instead of `Text` to save 2x space
+in a (false) hope that their data would never happen to contain anything non-ASCII.
+Backing `Text` by UTF-8 removes this source of inefficiency, reduces memory consumption
+up to 2x and promote more convergence to use `Text` for all stringy things
+(as opposed to binary data).
+
+Modern computer science research is focused on developing faster algorithms
+for UTF-8 data, e. g., an ultra-fast JSON decoder [simdjson](https://github.com/simdjson/simdjson). There is much less work on (and demand for) UTF-16 algorithms.
+Switching `text` to UTF-8 will open us a way to accomodate and benefit
+from future developments in rapid text processing.
 
 The importance of UTF-16 to UTF-8 transition was recognised long ago, and at least
-two attempts has been made:
+two attempts have been made:
 [in 2011](https://github.com/jaspervdj/text/tree/utf8) and five years later
 [in 2016](https://github.com/text-utf8/text-utf8). Unfortunately, they did not get
 merged into main `text` package. Today, five more years later it seems suitable
@@ -108,14 +116,13 @@ to make another attempt.
 
 -   Ensure stakeholders (e.g. GHC, Cabal, Stack, boot libs) have ample time to migrate and address any bugs.
 
--   Implementation should not significantly alter the performance characteristics of the base `text` library within some tolerance
-    threshold.
+-   Performance satisfies targets listed below in "Performance impact" section.
 
 # People
 
 -   Performers:
 
-    -   Leader: Andrew Lelechenko (bodigrim)
+    -   Leader: Andrew Lelechenko (Bodigrim)
 
     -   Support: Emily Pillmore (emilypi)
 
@@ -172,6 +179,73 @@ possible. This candidate should be shared publicly and loudly.
 **Implementation:**
 
 -   TBD: There is a straightforward implementation, but this one is left up to Andrew for comment.
+
+**Performance impact**
+
+A common misunderstanding is that switching to UTF-8 makes everything twice smaller and
+twice faster. That's not quite so.
+
+While UTF-8 encoded English text is twice smaller than UTF-16,
+this is not exactly true even for other Latin-based languages, which frequently
+use vowels with diacritics. For non-Latin scripts (Russian, Hebrew, Greek)
+the difference between UTF-8 and UTF-16 is almost negligible: one saves on
+spaces and punctuation, but letters still take two bytes. On a bright side, programs rarely
+transfer sheer walls of text, and for a typical markup language (JSON, HTML, XML),
+even if payload is non-ASCII, savings from UTF-8 easily reach \~30%.
+
+As a Haskell value, `Text` involves a significant constant overhead: there is
+a constructor tag, then offset and length, plus bytearray header and length.
+Altogether 5 machine word = 40 bytes. So for short texts, even if they are ASCII only,
+difference in memory allocations is not very pronounced.
+
+Further, traversing UTF-8 text is not necessarily faster than UTF-16. Both are
+variable length encodings, so indexing a certain element requires parsing everything
+in front. But in UTF-16 there are only two options: a code point
+takes either 2 or 4 bytes, and the vast majority of frequently used characters are
+2-byte long. So traversing UTF-16 keeps branch prediction happy. Now with UTF-8
+we have all four options: a code point can take from 1 to 4 bytes, and most non-English
+texts constantly alternate between 1-byte (e. g., spaces) and 2-byte characters.
+Having more branches and, more importantly, bad branch prediction is a serious penalty.
+This to a certain degree is mitigated by better cache locality.
+
+Existing `text` benchmarks are arguably favoring UTF-16 encoding: most of them are huge
+walls of Russian, Greek, Korean, etc. texts without any markup. So encoding them in UTF-8
+does not save any space, but we have to pay extra for more elaborate encoding. Our goal
+here is nevertheless to stay roughly on par with existing implementation.
+
+Benchmarks for `decodeUtf8` / `encodeUtf8` should improve significantly by virtue
+of avoiding conversion between UTF-8 and UTF-16 conversion.
+Fast validation of UTF-8 is not a trivial task, but we intend to employ
+[`simdjson::validate_utf8`](https://arxiv.org/pdf/2010.03090.pdf) for this task.
+
+Another important aspect of `text` performance is fusion.We are finalising
+an `inspection-testing`-based [test suite](https://github.com/haskell/text/pull/337) to check that
+pipelines, which used to fuse before, are fusing after UTF-8 transition as well.
+Fusion is incredibly fragile matter: for example, of 100 tests, which fuse in GHC 8.10.4,
+40 do not fuse in GHC 9.0.1, 30 do not fuse in GHC 8.4.4, etc. In such environment we cannot
+bet on retaining all fusion capabilities, but we aim to thoroughly investigate
+and explain all regressions.
+
+We expect that switching to UTF-8 will be beneficial for client of `text`, both
+libraries and applications. They'll be able to save memory for storage,
+save time on encoding/decoding inputs and outputs, use state-of-the-art text algorithms,
+developed for UTF-8. Parsers often benefit from UTF-8 encoding, because if a grammar
+does not have specific rules for non-ASCII characters (which is most often the case),
+parser can operate on a `ByteArray` without bothering about multibyte encodings at all.
+
+We will seek clients' feedback as early as possible, and will act on it, if it arrives
+before the end of the project. However, since our clients are external actors, often
+unpaid volunteers, we cannot expect them to provide feedback by the given date.
+Thus to keep targets of this project time-bound, we cannot include a goal
+of waiting for an approval of indefinite number of parties for indefinitely long.
+Such goal or sentiment, in our opinion, made a significant contribution into failure
+of two previous attempts.
+
+To sum up:
+
+* `decodeUtf8` and `encodeUtf8` become at least 2x faster.
+* Geometric mean of existing benchmarks (which favor UTF-16) decreases.
+* Fusion (as per our test suite) does not regress beyond at most several cases.
 
 **Stakeholders:**
 
--- a/proposals/002-text-utf-default.md
+++ b/proposals/002-text-utf-default.md
@@ -90,7 +90,7 @@ break fusion. Performance should not significantly diverge from the
 existing UTF-16 text package. There will be an expected change to the
 exposed Text internals, in which case, breakage should be assessed by
 circulating a git commit reference to a release candidate as soon as
-possible. This candidate should be sourced publicly and loudly.
+possible. This candidate should be shared publicly and loudly.
 
 **Implementation:**
 
--- a/proposals/002-text-utf-default.md
+++ b/proposals/002-text-utf-default.md
@@ -110,7 +110,7 @@ to make another attempt.
 
 # Goals
 
--   Solicit feedback from community members and industry to gauge appetite for such a change.
+-   Ensure that the new design of `text` and its migration strategy accommodates the needs of community members and industry
 
 -   Provide an implementation, migration, and delivery plan for changing the default encoding of `text` from UTF-16 to UTF-8.
 
--- a/proposals/002-text-utf-default.md
+++ b/proposals/002-text-utf-default.md
@@ -45,9 +45,9 @@ a character (as of Unicode 13.0). Now comes a tricky part: Unicode defines how t
 characters to code points (basically, integers), but how would you serialise lists of
 integers to bytes? The simplest encoding is just allocate 32 bits (4 bytes) per code
 point, and write them one by one. This is UTF-32. Its main benefit is that since
-all code points take the same size, so you can still index characters in a constant time.
+all code points take the same size, you can still index characters in a constant time.
 However, memory requirements are 4x comparing to ASCII, and in a world of ASCII and UCS-2
-there was little appetite to embrace one more, completely new encoding.
+there was little appetite to embrace one more incompatible encoding.
 
 Next option on the list is to encode some code points as 2 bytes and some others,
 less lucky ones, as 4 bytes. This is UTF-16. This encoding allowed to retain a decent
@@ -60,7 +60,7 @@ But once we abandon requirement of constant indexing, even better option arises.
 encode first 128 characters as 1 byte, some others as 2 bytes, and the rest as 4 bytes.
 This is UTF-8. The killer feature of this encoding is that it's fully backwards compatible
 with ASCII. This meant that all existing ASCII documents were automatically valid UTF-8
-documents as well, and that 50-years-old executables could often parse UTF08 data without
+documents as well, and that 50-years-old executables could often parse UTF-8 data without
 knowing a bit about it. This property appeared so important that in a modern environment
 the vast majority of data is stored and sent between processes in UTF-8 encoding.
 
@@ -72,13 +72,33 @@ To sum up:
 
 # Motivation
 
--   UTF-16 by default requires that all Text values pay a premium for serialization. Arguably, the performance impact of Text is flipped
-    upside-down: most text is UTF-8, and Haskell devs pay an undue cost when working with the wrong default.
+`text` is a standard Haskell library for Unicode strings. Internally it stores
+Unicode code points in UTF-16, so any character takes either 2 or 4 bytes.
+In a modern enviroment this is a suboptimal choice: usually data
+is stored (e. g., on a disc or in DB) and tranferred between agents (e. g., via web)
+in UTF-8 encoding. So `text` needs to convert (UTF-8 to UTF-16) all inputs and usually
+ends up converting outputs as well (this time UTF-16 to UTF-8).
 
--   UTF-8 is the industry standard and by far the most common text encoding, with roughly 97% of web pages existing in UTF-8. The
-    existing UTF-16 default imposes an additional hurdle to working with the vast majority of web content on earth.
+Even within Haskell ecosystem UTF-16 is rarely used
+for interprocess communication or as a component of binary formats.
+The very `instance Binary Text` serializes `Text` in UTF-8 encoding.
 
--   Many systems in Haskell are UTF-8 by default (e.g. Haddock)
+If we switch the internal representation of `Text` from UTF-16 to UTF-8,
+all such conversions would be made redundant and we'll be able just check that
+a `ByteString` is a valid UTF-8 (which is most often the case) and copy it into `Text`.
+If in future (see an upcoming "Unifying vector-like types" proposal) `ByteString`
+is backed by unpinned memory, we'd be able to eliminate copying entirely.
+
+`Text` is also often used in contexts, which involve mostly ASCII characters.
+For such applications storing data in UTF-8 means using up to 2x less space,
+which could be important to reduce memory pressure.
+
+The importance of UTF-16 to UTF-8 transition was recognised long ago, and at least
+two attempts has been made:
+[in 2011](https://github.com/jaspervdj/text/tree/utf8) and five years later
+[in 2016](https://github.com/text-utf8/text-utf8). Unfortunately, they did not get
+merged into main `text` package. Today, five more years later it seems suitable
+to make another attempt.
 
 # Goals
 
--- a/proposals/002-text-utf-default.md
+++ b/proposals/002-text-utf-default.md
@@ -97,7 +97,9 @@ up to 2x and promote more convergence to use `Text` for all stringy things
 (as opposed to binary data).
 
 Modern computer science research is focused on developing faster algorithms
-for UTF-8 data, e. g., an ultra-fast JSON decoder [simdjson](https://github.com/simdjson/simdjson). There is much less work on (and demand for) UTF-16 algorithms.
+for UTF-8 data, e. g., an ultra-fast JSON decoder
+[simdjson](https://github.com/simdjson/simdjson).
+There is much less work on (and demand for) UTF-16 algorithms.
 Switching `text` to UTF-8 will open us a way to accomodate and benefit
 from future developments in rapid text processing.
 
@@ -110,7 +112,7 @@ to make another attempt.
 
 # Goals
 
--   Ensure that the new design of `text` and its migration strategy accommodates the needs of community members and industry
+-   Ensure that the new design of `text` and its migration strategy accommodates the needs of community members and industry.
 
 -   Provide an implementation, migration, and delivery plan for changing the default encoding of `text` from UTF-16 to UTF-8.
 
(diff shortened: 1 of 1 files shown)


proposals/002-text-utf-default.md
```diff
@@ -0,0 +1,156 @@
...
+
+# Motivation
+
+-   UTF-16 by default requires that all Text values pay a premium for serialization. Arguably, the performance impact of Text is flipped
+    upside-down: most text is UTF-8, and Haskell devs pay an undue cost when working with the wrong default.
+
+-   UTF-8 is the industry standard and by far the most common text encoding, with roughly 97% of web pages existing in UTF-8. The
+    existing UTF-16 default imposes an additional hurdle to working with the vast majority of web content on earth.
+
+-   Many systems in Haskell are UTF-8 by default (e.g. Haddock)
```
goldfirere, line 24: How do such systems have this default? Do these systems avoid `text` entirely? Is there an alternative to `text` that uses UTF-8?

proposals/002-text-utf-default.md
```diff
@@ -0,0 +1,156 @@
...
+
+# Goals
+
+-   Solicit feedback from community members and industry to gauge appetite for such a change.
```
goldfirere, line 28: From the posting of the proposal, it sounds like this goal is already met. Is there more to do here?

Also, to me, this isn't a goal, but an action step. The goal might be "Ensure that the new design of `text` and its migration strategy accommodates the needs of community members and industry", which would be supported by the action step written here.

cartazio:
looks good to me

proposals/002-text-utf-default.md
```diff
@@ -0,0 +1,156 @@
...
+
+-   Provide an implementation, migration, and delivery plan for changing the default encoding of `text` from UTF-16 to UTF-8.
+
+-   Ensure stakeholders (e.g. GHC, Cabal, Stack, boot libs) have ample time to migrate and address any bugs.
+
+-   Implementation should not significantly alter the performance characteristics of the base `text` library within some tolerance
```
goldfirere, line 34: What tolerance threshold? 10x is "some" threshold, but probably not the one you want. I wouldn't see a violation of this threshold as a standalone reason to declare failure, but if we say the new implementation should be, say, within 5% of the performance of the old one, and it's an 8% regression, it would mean at least advertising this downside and seeking more community input.

proposals/002-text-utf-default.md
```diff
@@ -0,0 +1,156 @@
...
+
+        -   Dan Cartwright (chessai)
+
+        -   Callan McGill (boarders)
+
+-   Stakeholders:
+
+    -   Edward Kmett: has been vocal about his use of `text-icu` and requires it not be broken.
+
+-   Ben Gamari: integration with GHC
```
goldfirere, line 61: Should these two entries be indented under Stakeholders?

proposals/002-text-utf-default.md
```diff
@@ -0,0 +1,156 @@
...
+
+-   Update testing and performance benchmarks to make use of `inspection-testing` to ensure fusion is not broken in
+    subsequent UTF-8 related changes.
+
+An MVP should completely preserve standard user-facing API, and not
+break fusion. Performance should not significantly diverge from the
+existing UTF-16 text package. There will be an expected change to the
+exposed Text internals, in which case, breakage should be assessed by
+circulating a git commit reference to a release candidate as soon as
+possible. This candidate should be sourced publicly and loudly.
```
goldfirere, line 93:
```suggestion
possible. This candidate should be shared publicly and loudly.
```

proposals/002-text-utf-default.md
```diff
@@ -0,0 +1,156 @@
...
+# Timeline
+
+We expect that this project will take roughly 6 months to fully
+complete: 3-4 months to complete the code implementation, performance
+testing, and unit testing, another 1-2 months to integrate with
+stakeholders and diagnose any potential issues with the migration.
+
+**Preparation:**
+
+Using the HVR's existing [`text-utf8`](https://github.com/text-utf8) as
```
ketzacoatl, line 77: Is this link incomplete?

proposals/002-text-utf-default.md
```diff
@@ -0,0 +1,156 @@
...
+there will be some help that needs to be given in terms of bumping
+Hackage bounds since this migration will be a major version bump. HF
+will need to coordinate with the Hackage Trustees to help move along
+packages that go out of date.
+
+# Deliverables
+
+-   text-2.0.0.0, which will provide a UTF-8 encoding for Text as a default for all versions going forward.
+
+-   A `text-utf16` package, which is a preservation of the current UTF-16 encoded text, for backwards compatibility.
```
goldfirere, line 117: In reaching out to the community, is there demand for this support? If so, is there a reason to make this its own package, and not just e.g. a different datatype exported from the same `text` package?

proposals/002-text-utf-default.md
```diff
@@ -0,0 +1,156 @@
...
+-   Updates to the Text Haddocks that reflect the UTF-8 changes
+
+-   Announcements and updates across all Haskell channels covering the following:
+    -   Significant dates and milestones
+
+    -   Expected code impact
+
+    -   Release candidates
+
+    -   Delivery
```
goldfirere, line 128: Can we add performance impact analysis to this list?

proposals/002-text-utf-default.md
```diff
@@ -0,0 +1,156 @@
...
+
+-   Migration instructions and design documentation
+
+# Outcomes
+
+-   Addresses a recognized need and want from both Industry and the Haskell Community
+
+-   Better UX for Haskell's text story
```
goldfirere, line 136: How so? This bullet implies a change to the API of `text`, but other commentary here says that the interface will not change (except for exported internals).

proposals/002-text-utf-default.md
```diff
@@ -0,0 +1,156 @@
...
+
+-   Establishes the ability for Haskell Foundation to Get Things Done that were previously blocked for 10+ years in the community.
```
goldfirere, line 138: 🎉

proposals/002-text-utf-default.md
```diff
@@ -0,0 +1,156 @@
...
+-   Better interoperability and web story
+
+# Risks
+
+-   HF must minimize the cost to migrate, or people will just get mad (and rightfully so).
+
+-   Text-icu will need a bespoke UTF-8 conversion function. In general, the Unicode story must be tracked and made sure it will not break.
+    -   Recommendation: make this a high-priority deliverable when project planning
+
+-   The old UTF-16 text package will need to be preserved, and will require a maintainer.
```
goldfirere, line 149: Why?

proposals/002-text-utf-default.md
emilypi: Possible. I'll fix it.

proposals/002-text-utf-default.md
```diff
@@ -0,0 +1,156 @@
...
+
+-   Performance expectations should be managed: UTF-8 text is \*not\* a panacea. It will not result in a 2x or even significant
```
goldfirere, line 151: This surprises me -- and in fact calls into question the original motivation.

As I read the proposal as written, I see that this proposal "should completely preserve standard user-facing API" and "not result in ... significant performance increase in many cases", while "performance may regress in some programs". If there is no change to the API and minimal expected change in performance, then what is the concrete benefit here? I do see that people have been clamoring for this change for some time -- that's great. But I do not understand from this proposal what concrete benefit this proposal brings. The best I can extract from this proposal is that it stops people from saying "Haskell is so backward -- it uses UTF-16 as its most common performant text encoding!". Maybe that's the underlying reason? It is good not to be backward, but we should be clear on what our motivation really is, so that we can evaluate later whether we've met our goals.

proposals/002-text-utf-default.md
```diff
@@ -0,0 +1,156 @@
...
+
+-   Solicit feedback from community members and industry to gauge appetite for such a change.
+
+-   Provide an implementation, migration, and delivery plan for changing the default encoding of `text` from UTF-16 to UTF-8.
+
+-   Ensure stakeholders (e.g. GHC, Cabal, Stack, boot libs) have ample time to migrate and address any bugs.
+
+-   Implementation should not significantly alter the performance characteristics of the base `text` library within some tolerance
+    threshold.
+
```
goldfirere, line 36: Are there any performance improvement goals? For example: a 10% speedup of reading HTML delivered off the web (presumably in UTF-8 format).

goldfirere, line 36: Are there any API stability goals? I know this is mentioned elsewhere, but it would be nice to define here that, say, a concrete goal is that all the functions exported from `Data.Text` remain unchanged in their types and semantics (but may change in performance characteristics). I have no idea whether that's the right goal, but if stability is a goal of the project, this is the place to put it.

proposals/002-text-utf-default.md
```diff
@@ -0,0 +1,156 @@
...
+-   Better UX for Haskell's text story
+
+-   Establishes the ability for Haskell Foundation to Get Things Done that were previously blocked for 10+ years in the community.
+
+-   Better interoperability and web story
+
+# Risks
+
+-   HF must minimize the cost to migrate, or people will just get mad (and rightfully so).
```
goldfirere, line 144: Where do we expect these costs to arise? If the API remains stable, migration costs should be minimal (zero?) for many users. There might be performance impacts, but these are very hard to mitigate, as (I think) a perf impact would have to be measured by each consumer who cares about perf, on their own particular use case.

goldfirere (commented):
It's so great to have all these details in one place. Textual encoding is not an area I know much about, and so I haven't been able to follow this track well. Now, I see it all laid out. Thank you!

While I have a bunch of comments here, please do not see this as trying to defeat this proposal. I just want the text to be clear about the concrete wins for Haskell that we should expect after this is complete.

proposals/002-text-utf-default.md
ketzacoatl: I might be totally confused and wrong about this, but I was under the impression the changes outlined in this proposal (amongst other results) open the door / hallway towards improving the situation with `String`/`Text`/`ByteString` and the prolific use of `String` in our "core libraries" (despite the prolific recommendation to use `Text`/`ByteString` instead of `String`.

proposals/002-text-utf-default.md
cartazio: text using utf 16,is the IN memory rep, not the choice of encoding it reads from the world/writes back out. the text package provides translation encoders/decoders

proposals/002-text-utf-default.md
davean: It halves the memory usage in the average case, using 8 instead of 16 for most cases. This can help, or harm performance. It can increase or decrease memory usage. It varies by the workload.

proposals/002-text-utf-default.md
torgeirsh: Perhaps it could be clarified what "are UTF-8" means? Does it mean having a UTF-8 in-memory representation by other means than the text package, or that they use text's current UTF-16 representation, but usually input/output UTF-8?

proposals/002-text-utf-default.md
cartazio: continuous feedback is how we avoid waterfall method:)

proposals/002-text-utf-default.md
cartazio: I think this is largely about managing expectations, because eg making sure changes in eg code size dont create differences in how inlining and fusion operate on otherwise observationaly equal programs is quite reasonably out of scope for this project (text is very dependent on good quality fusion optimization for good perf by most users, but making fusion rules more robust is out of scope for this work)

proposals/002-text-utf-default.md
jberryman: I also wonder why. Certainly libraries that use `text` internals (like `hashable`) will need to be re-written or have a new CPP case to handle the old encoding, but I don't immediately see the utility of `old-text`

proposals/002-text-utf-default.md
phadej: There is potential for performance boost, as most on the wire text is UTF8
nowadays, so it only needs to be validated not necessarily copied.

However, if data is read as `ByteString` (backed by `ForeignPtr`) (as it often is)
then we still would need to copy to `ByteArray#` backing up `Text`.

So my gut feeling is that performance gains will arise mostly from reduced memory pressure.
That said, for e.g. Russian or Greek text the performance won't be any better. Or any other script which is mostly outside of ASCII (note, not even 256 first codepoints) range.

E.g. in finnish which is mostly ASCII but has common öä:

```
-rw-rw-r--  1 phadej phadej   597661 May 13 22:45 kalevala.txt
-rw-rw-r--  1 phadej phadej  1150892 May 13 22:46 kalevala16.txt
```

the UTF-8 version is 51% of UTF-16 text.

However with Russian (this is a word list file)

```
-rw-rw-r--  1 phadej phadej 36530302 May 13 22:50 russian16.txt
-rw-rw-r--  1 phadej phadej 34993984 May 13 22:50 russian8.txt
```

the UTF-8 version is 95% of UTF-16 (the saving comes from line feeds and space characters). Almost the same.

---

I.e. any performance changes are not given.
E.g. Z-Data performance gains are (as far as I can tell) from copy avoidance, but `ForeignPtr` vs. `ByteArray#` ruin that chance.

phadej:
The proposal should mention what happens with `text-short`: https://hackage.haskell.org/package/text-short

If my memory is correct, HVR intended to add `ShortText` directly into `text` but bos (who was still around then) didn't like that idea.

However, for use as "text symbols", e.g. representing identifiers in the programming language `ShortText` is IMO more suitable, as it doesn't have slicing - which you don't really need for the identifiers. Remember that offset+length is 16 bytes on 64bit system, which is a lot for every identifier.

I could argue that e.g. `aeson` might benefit from switching from `Text` to `ShortText` in the representation of JSON (at least object keys, maybe even all strings). I'm sure users would hate me for that change, but I'm also quite sure `aeson` itself could be faster with that change!

`text-short` is already using UTF-8.

proposals/002-text-utf-default.md
phadej: It is correct. https://github.com/text-utf8 is a GitHub organization.

proposals/002-text-utf-default.md
goldfirere: Thanks @cartazio for the explanation. Yes, that makes sense. So this point is saying that these tools require conversion on (de)serialization, implying that the status quo for `text` makes these tools less performant. It might good to spell that out.

proposals/002-text-utf-default.md
jberryman: I agree with Richard, and think the motivation section is not particularly clear...

> UTF-16 by default requires that all Text values pay a premium for serialization. Arguably, the performance impact of Text is flipped upside-down: most text is UTF-8, and Haskell devs pay an undue cost when working with the wrong default.

This makes complete intuitive sense to me! ...but has this hypothesis been validated? If it's not true then it's not clear what reasons remain for this change.

If measurable performance improvement is what will determine whether this is considered a success and merged (which it sounds like it should), have stakeholders agreed on a set of benchmarks ahead of time?

The other points don't seem like "motivation" to me; e.g. there's no barrier to using `text` with UTF-8 because it's already been written.

proposals/002-text-utf-default.md
Lysxia: > However, if data is read as `ByteString` (backed by `ForeignPtr`) (as it often is)
then we still would need to copy to `ByteArray#` backing up `Text`.

Why must `Text` be backed by `ByteArray#`? What are the cons of having a `Text` where decoding from UTF-8 makes no copy?

proposals/002-text-utf-default.md
goldfirere: All I'm saying is that we should name a specific goal threshold for how much regression we're willing to tolerate before we take a step back and wonder whether this was a good idea.

proposals/002-text-utf-default.md
phadej: In theory the full matrix of
- UTF8, UTF16, UTF32 (LE, BE)
- sliced, non-sliced
- ByteArray#, pinned/unpinned, ForeignPtr backed

can be useful (c.f. avoiding copying etc.).

(for example, some people have been experimenting with parsing UTF32 backed text - it has random access, so it's great when inputs are of human size, re-encoding cost may be won in faster parsing).

phadej:
Also, I'd like to have comments why we think UTF8 is better, when e.g.

- JavaScript uses UTF16: https://developer.mozilla.org/en-US/docs/Web/API/DOMString,
- V8 story is more complicated: https://stackoverflow.com/questions/40512393/understanding-string-heap-size-in-javascript-v8
   -  Latin1 strings
   - UTF16 strings
   - sliced strings (c.f. `Text` vs `ShortText`)
   - ...
   - which means they have opaque `Text` which is smart in what it's actually backed up by - not commiting to a single choice!

- Java String is UTF16, https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html

>  String represents a string in the UTF-16 format in which supplementary characters are represented by surrogate pairs

proposals/002-text-utf-default.md
goldfirere: Can we translate this understanding into a goal of this work? For example, "The geometric mean number of allocations of tests in the xyz benchmark suite goes down by 40%, with no regression above 20%." That would be fantastic! Of course, we might not meet the goal. But @davean is suggesting that the common case would halve memory usage, and that (to me) is in conflict with "not result in ... significant performance increase in many cases". So which is it? The "halve memory usage" seems very plausible -- so we should state that as a goal and then measure to see whether we've met that goal.

proposals/002-text-utf-default.md
goldfirere: Thanks @phadej for this explanation. I'm not surprised here -- performance is always variable. But there still can be performance goals. For example: "Reduction of 50% of memory usage on ASCII text". We're restricting the input here -- ASCII text only -- but still stating an admirable goal. I bet we can do even better, though, by stating performance goals over averages (while limiting regressions of any particular test).

The underlying reason we want this proposal is to improve performance of `text` (I think), so it seems short-sighted not to state a goal of improving performance of `text`, suitably qualified to account for reality.

phadej: @Lysxia, one can argue in different directions. Why `ByteString` have to be backed by `ForeignPtr`. The only good answer I know is to support `mmap`, as we cannot create `ByteArray#` from external memory (why can't we?).

OTOH, it's extremely tempting to be able to `mmap` textual file into (strict) `Text`. That **will** save time.

phadej: @goldfirere the ASCII text is very West centric view. Note, we talk about `Text`, which is used for processing text as in books, chat messages (!!!) etc.

"programming languages" may be important, but in "real world" that is still a niche use case.

phadej: Recall the representation:
- header 8 bytes
- offset + length: 2x8bytes = 8bytes
- payload: at least 2 bytes per codepoint

Saving 50% memory usage is technically impossible. To get close you need to have big strings. (Like whole comments e.g.) and to benefit from slicing you need overlapping slices.

Slices are useful so `drop 1` etc. are fast - they don't save memory.

---

What I try to say, without success, is that if you are concerned with performance then there cannot be a single `Text` representation which is just best. 

A simple example is that people don't use `Data.Text.Builder`, but just concatenate strict  `Text` values (in logging). That is just bad, O(n^2) copying. Where simple `"foo" ++ show aNumber ++ "bar" :: String` is actually not that bad!

goldfirere: I agree that ASCII text is very limiting. So perhaps that's the wrong metric. But we need to identify *some* goal here, otherwise we won't know whether we've been successful.

As it stands, the proposal's net effect is "this proposal will change performance of `text`". I don't doubt that performance will change. But if that's the best net effect we can muster, it doesn't seem, to me, that this proposal is worth investing time and energy behind. On the other hand, if we can state a goal of improving (average) performance across a range of inputs, that sounds exciting. Maybe we'll meet the goal, and maybe we won't, but having a goal at least frames the discussion so we can evaluate, later, whether merging is a good idea.

Maybe more simply: the fact that performance is variable across inputs is not (to me) a reason not to measure performance.

goldfirere: > What I try to say, without success, is that if you are concerned with performance then there cannot be a single Text representation which is just best.

I think this has been successful. I appreciate this, and I appreciate that, whatever we do, some applications of `Text` will get faster and some will get slower. That's all good. But I do hope that we can expect an improvement **on average** (for a suitably fair definition of average).

coderfromhere:
> Also, I'd like to have comments why we think UTF8 is better, when e.g.

This link might be useful: http://utf8everywhere.org

proposals/002-text-utf-default.md
jberryman: In addition to stakeholders agreeing on a set of benchmarks worthy of both _optimizing against_ and _testing against for regressions_, I think @phadej 's comments make it clear that micro-benchmarks won't be sufficient... you'll need to measure the entire life-cycle: serialization, deserialization, additional copying, effects on memory and runtime of sharing/slicing, FFI, etc.

proposals/002-text-utf-default.md
cartazio: ok, "commiting to a decision threshold where we should look at what is failing and why to learn from and fix" sounds like a good idea

chrisdone:
I can add an example motivation. The aeson package parses from a ByteString assumed to be UTF-8. Then for each string you get a Text. If you’re using the fastest parsers you can, they will typically be e.g. attoparsec or flatparse which work on a ByteString. These often take advantage of the byte by byte format, using fast memchr and such functions from C. In order to use such parsers, you’re going from UTF-8, to UTF-16 —doubling the size, then you’ll have to again encode back into UTF-8. I’ve measured the performance of encodeUtf8 and found it to be quite small. About 10 microseconds per kilobyte of Unicode on my laptop. So doing this extra step is “OK”, but that depends on your app. I’ll attach my benchmarks here.

chrisdone:
```haskell
sampleUnicode :: Text
sampleUnicode = "! \" # $ % & ' ( ) * + , - . / 0 1 2 3 4 5 6 7 8 9 : ; < = > ? @ A B C D E F G H I J K L M N O P Q R S T U V W X Y Z [ \\ ] ^ _ ` a b c d e f g h i j k l m n o p q r s t u v w x y z { | } ~ ¡ ¢ £ ¤ ¥ ¦ § ¨ © ª « ¬ ® ¯ ° ± ² ³ ´ µ ¶ · ¸ ¹ º » ¼ ½ ¾ ¿ À Á Â Ã Ä Å Æ Ç È É Ê Ë Ì Í Î Ï Ð Ñ Ò Ó Ô Õ Ö × Ø Ù Ú Û Ü Ý Þ ß à á â ã ä å æ ç è é ê ë ì í î ï ð ñ ò ó ô õ ö ÷ ø ù ú û ü ý þ ÿ Ā ā Ă ă Ą ą Ć ć Ĉ ĉ Ċ ċ Č č Ď ď Đ đ Ē ē Ĕ ĕ Ė ė Ę ę Ě ě Ĝ ĝ Ğ ğ Ġ ġ Ģ ģ Ĥ ĥ Ħ ħ Ĩ ĩ Ī ī Ĭ ĭ Į į İ ı Ĳ ĳ Ĵ ĵ Ķ ķ ĸ Ĺ ĺ Ļ ļ Ľ ľ Ŀ ŀ Ł ł Ń ń Ņ ņ Ň ň ŉ Ŋ ŋ Ō ō Ŏ ŏ Ő ő Œ œ Ŕ ŕ Ŗ ŗ Ř ř Ś ś Ŝ ŝ Ş ş Š š Ţ ţ Ť ť Ŧ ŧ Ũ ũ Ū ū Ŭ ŭ Ů ů Ű ű Ų ų Ŵ ŵ Ŷ ŷ Ÿ Ź ź Ż ż Ž ž ſ ƀ Ɓ Ƃ ƃ Ƅ ƅ Ɔ Ƈ ƈ Ɖ Ɗ Ƌ ƌ ƍ Ǝ Ə Ɛ Ƒ ƒ Ɠ Ɣ ƕ Ɩ Ɨ Ƙ ƙ ƚ ƛ Ɯ Ɲ ƞ Ɵ Ơ ơ Ƣ ƣ Ƥ ƥ Ʀ Ƨ ƨ Ʃ ƪ ƫ Ƭ ƭ Ʈ Ư ư Ʊ Ʋ Ƴ ƴ Ƶ ƶ Ʒ Ƹ ƹ ƺ ƻ Ƽ ƽ ƾ ƿ ǀ ǁ ǂ ǃ Ǆ ǅ ǆ Ǉ ǈ ǉ Ǌ ǋ ǌ Ǎ ǎ Ǐ ǐ Ǒ ǒ Ǔ ǔ Ǖ ǖ Ǘ ǘ Ǚ ǚ Ǜ ǜ ǝ Ǟ ǟ Ǡ ǡ Ǣ ǣ Ǥ ǥ Ǧ ǧ Ǩ ǩ Ǫ ǫ Ǭ ǭ Ǯ ǯ ǰ Ǳ ǲ ǳ Ǵ ǵ Ǻ ǻ Ǽ ǽ Ǿ ǿ Ȁ ȁ Ȃ ȃ ɐ ɑ ɒ ɓ ɔ ɕ ɖ ɗ ɘ ə ɚ ɛ ɜ ɝ ɞ ɟ ɠ ɡ ɢ ɣ ɤ ɥ ɦ ɧ ɨ ɩ ɪ ɫ ɬ ɭ ɮ ɯ ɰ ɱ ɲ ɳ ɴ ɵ ɶ ɷ ɸ ɹ ɺ ɻ ɼ ɽ ɾ ɿ ʀ ʁ ʂ ʃ ʄ ʅ ʆ ʇ ʈ ʉ ʊ ʋ ʌ ʍ ʎ ʏ ʐ ʑ ʒ ʓ ʔ ʕ ʖ ʗ ʘ ʙ ʚ ʛ ʜ ʝ ʞ ʟ ʠ ʡ ʢ ʣ ʤ ʥ ʦ ʧ ʨ"
```
```haskell
bgroup
  "encodeUtf8"
  [ env
    (pure (T.replicate i sampleUnicode))
    (\t ->
       bench
         ("T.encodeUtf8: " ++ show (i * T.length sampleUnicode) ++ " chars")
         (whnf T.encodeUtf8 t))
  | i <- [1, 10, 100]
  ]
```

Numbers:
```
benchmarking encodeUtf8/T.encodeUtf8: 1065 chars
time                 1.059 μs   (1.050 μs .. 1.070 μs)
                     0.999 R²   (0.999 R² .. 1.000 R²)
mean                 1.058 μs   (1.049 μs .. 1.070 μs)
std dev              34.92 ns   (27.00 ns .. 49.58 ns)
variance introduced by outliers: 46% (moderately inflated)

benchmarking encodeUtf8/T.encodeUtf8: 10650 chars
time                 9.998 μs   (9.937 μs .. 10.07 μs)
                     0.999 R²   (0.999 R² .. 1.000 R²)
mean                 10.03 μs   (9.958 μs .. 10.15 μs)
std dev              313.1 ns   (237.3 ns .. 478.0 ns)
variance introduced by outliers: 37% (moderately inflated)

benchmarking encodeUtf8/T.encodeUtf8: 106500 chars
time                 100.0 μs   (98.89 μs .. 101.4 μs)
                     0.999 R²   (0.998 R² .. 1.000 R²)
mean                 99.66 μs   (98.81 μs .. 100.6 μs)
std dev              2.918 μs   (2.402 μs .. 3.566 μs)
variance introduced by outliers: 27% (moderately inflated)
```

So about 1us per kilobyte. Pretty consistently, for a small subset of Unicode. For Chinese, the cost would be higher, but I haven't tested.

Anyway, my point was to just bring some numbers to give people ballpark figures. 

I'd love to remove this redundant decode/encode step and fully support this initiative.

chrisdone:
Apologies if this was spelled out somewhere I haven't seen: but was there a consideration and decision made on having a package flag like "-futf8" or "-futf16" - allowing the authors to pick based on benchmarking their app?

Web apps will be naturally better performing as most of everything web is UTF-8, whereas anything talking to ODBC or Windows may benefit from direct UTF-16 use. If I could just pick with a flag in my stack.yaml for the text package, that'd mean I at least get to make a decision rather than having it hard-coded into the ecosystem like it is in Rust. :thinking: 

Any opinions on that?

proposals/002-text-utf-default.md
emilypi: Yeah, confirmed that the link is not broken. Thanks for explaining to @ketzacoatl, @phadej

proposals/002-text-utf-default.md
emilypi: Thanks all - I'll clarify what we mean here.

phadej:
> Any opinions on that?

It will be tricky. It is possible to generate a header `text` will install so the downstream libraries may do things differently based on the encoding of `text`. Whether the maintainers will do that, we cannot know without asking. Whether they will even actively test non-default configuration is a very good question to ask.

phadej:
If `ghc`-the-library will ever depend on `text` directly, re-installing `text` will be impossible if `ghc` is not made reinstallable (which I doubt will happen before this change is implemented). So there is that "but" also.

proposals/002-text-utf-default.md
emilypi: This would be a good deliverable actually. The API *must* remain the same in order for migration to be as seamless as possible. Since we're building on `text-utf8`, this is actually already done for us for the most part, and we intend to keep it confluent with the existing `text` API. I'll add this.

proposals/002-text-utf-default.md
emilypi: This ties in with the above comment. We'll detail.

chrisdone:
Good points. Inclusion with GHC would be a good reason to not bother with a flag. Assuming it’ll be hard to build GHC packages for a long time.

You’re probably right about maintenance too. It’s a lot to ask to maintain text already. Just moving it to use UTF-8 internally seems more achievable and long-term sustainable. 👍

proposals/002-text-utf-default.md
gbaz: I don't think you should name a specific threshold. Rather, once the implementation is underway, and a sense of the tradeoffs is more apparent, _then_ a decision should be made, based on the facts at hand, not an arbitrary before-hand choice.

proposals/002-text-utf-default.md
gbaz: Why? Because there are many cases in which utf-16 _is_ the better choice -- that's why it was picked in the first place!

proposals/002-text-utf-default.md
gbaz: I think the argument is that there will be variable impact across a wide range of different applications, and therefore picking _beforehand_ a concrete metric is not the right approach.

If one is doing the "textbook scientific method" (that no scientists actually use) where you fix a hypothesis and then perform an experiment, then this is what you need to do. But that's not the way engineering works, and its not arguably even how lots of science works. When you do a drug trial, you statistically validate against the null hypothesis, but you don't fix a "amount of cancer cured" threshold before running the trial.

So I would say there is strong suspicion this will be helpful in certain large applications (but perhaps not others) and on measuring the impact after the fact, a decision can be made on what to do as the default.

Arguably, the fact that there isn't a single quantifiable metric to reduce everything to, but instead this range of tradeoffs and unknowns is partly what has stalled similar projects for so long. I'd say that there's sufficiently broad sentiment to try to do this that its worth it to first do it, then just see what happens, and then proceed based on that, and also now there will be a "definitive answer" as to why things are why they are, based on a significant investment of resources both in development and in benchmarking.

neilmayhew:
I'm not sure how relevant this is, but this project is very thought-provoking:

[simdjson : Parsing gigabytes of JSON per second](https://github.com/simdjson/simdjson)

As well as parsing JSON, it builds on earlier work using SIMD to validate UTF-8. The performance of both projects is astonishing, and the associated academic papers are very readable.

I wonder how possible it would be to make use of these libraries through FFI, and whether that would be a justification for using an FFI-friendly memory representation. If it were possible to read/mmap an entire file and validate the UTF-8 in a fraction of a second, I think that would be a big win.

proposals/002-text-utf-default.md
Bodigrim: As long as your grammar do not have rules specific for non-ASCII characters (which is most often the case), one can parse UTF8-encoded data as a `ByteString`, benefiting from random access. Same as UTF32, but in 4x less memory. Note that this is not the case for UTF16, since it is not compatible with ASCII.

proposals/002-text-utf-default.md
bgamari: Yes, this is true but In this case we should call the existing package not `old-text` but rather `text-utf16`.

proposals/002-text-utf-default.md
bgamari: Yes, but then we also need to define what we mean by "the `text` API".

proposals/002-text-utf-default.md
```diff
@@ -13,6 +15,61 @@ appetite for breakage, and what stakeholders would be affected by the
...
+
+Next option on the list is to encode some code points as 2 bytes and some others,
+less lucky ones, as 4 bytes. This is UTF-16. This encoding allowed to retain a decent
+backward compatibility with UCS-2, so, for instance, modern Java and JavaScript
+stick to UTF-16 as default internal representation. The biggest downside is that you
+no longer know beforehand, where *n*-th code point starts. One must *parse* all characters
+from the very beginning to learn which ones are 2-byte long and which are 4-byte long.
+
+But once we abandon requirement of constant indexing, even better option arises. Let's
+encode first 128 characters as 1 byte, some others as 2 bytes, and the rest as 4 bytes.
```
neilmayhew, line 60: Actually, U+0800 - U+FFFF are encoded as 3 bytes.

proposals/002-text-utf-default.md
emilypi: Why the qualifier? I see no reason why someone shouldn't be able to pick up `text-utf16` and continue maintaining it as the canonical UTF-16 text package. We don't want it to be the default, or perhaps be in `base`, but it's not a _bad_ package per se, as is implied by `old`.

proposals/002-text-utf-default.md
phadej: If you parse from bytes, you should use `ByteString` (or something with name "bytes" in it). Then `Text` encoding is irrelevant.

proposals/002-text-utf-default.md
phadej: @gbaz, I agree. But that should be written explicitly into the proposal in some way.

Otherwise when things are done, sunk cost etc and the change is pushed forward no matter what are result.

Experiments should be allowed to fail. Currently this is not setup as such experiment.

goldfirere: All I'm arguing for here is a clear metric by which we will evaluate whether this effort has been successful. Perhaps I have over-focused on percentages -- maybe it would instead be better to focus on the choice of benchmarks. In any case, I'm in support of looking at the results holistically later to decide whether this project has been successful. However, we need to decide the metric of evaluation (that is, the benchmarks) ahead of time. Otherwise, as @phadej says, we will be able to declare success no matter what happens.

Furthermore, I still think having goal metrics (not just chosen benchmarks) is a good idea. Despite the various claims here that it is impossible to get such goals, I've seen two plausible ones floated: @davean says (https://github.com/haskellfoundation/tech-proposals/pull/1#discussion_r632036534) that memory usage of UTF-8 should be roughly half that of UTF-16 for ASCII text. And @chrisdone describes (https://github.com/haskellfoundation/tech-proposals/pull/1#issuecomment-840870519) that the conversion takes 1us/kb, even providing a concrete test (thank you!). These observations can provide the basis of goals.

Here's another way to think about it: We, as a community, are choosing whether to invest in the idea presented here. At what point will our investment be worth it? If 1% is shaved off the running time of one application, but all others suffer a 10% regression, is this worth pursuing? (I imagine not.) If the geometric mean of a suite of applications shows no improvement, some applications improve by 2%, and a few regress by 20%, is it worth it?  (This is a harder one.) This thinking gives us a different way to phrase the goal: identify the minimum improvement we should expect to see in order to declare success. Hopefully, we'll far surpass it. Maybe we'll fall short -- but there might be mitigating reasons to accept the change anyway.

To connect most directly with @gbaz's argument: the development of the covid vaccine is about as experimental as you can get. Yet it still had a goal: in the US, the goal of 50% efficacy was frequently mentioned (where efficacy was well-defined in terms of double-blind trials and particular ways of measuring). I doubt anyone had any idea of whether that goal could be achieved -- but it was identified as a goal so that we could say, if a vaccine arrived with only 40% efficacy, that we should keep searching and not do a major campaign to get everyone vaccinated. We have, amazingly, far surpassed that goal with several highly effective vaccines. Just as I hope that the work proposed here will far surpass a modest goal we might set.

gbaz: " If 1% is shaved off the running time of one application, but all others suffer a 10% regression, is this worth pursuing? (I imagine not.)" Ok, but how will we know this until the library is written and we then test it against a bunch of stuff?

I don't think setting specific percentages at all helps here. One could state general expectations -- that it improves some things by a noticeable amount (likely memory-dependent things and ffi dependent things), and generally doesn't cause regressions elsewhere. The only real metric that I think matters is "the good outweighs the bad in a way that on the whole improves things for most users." 

Expectations of performance increase in any more detail are more _motivations_ than criteria for success.

goldfirere: > Ok, but how will we know this until the library is written and we then test it against a bunch of stuff?

We won't know this. That's the whole point: we state what we'd like to achieve and then try to achieve it. If we can't, that doesn't mean failure -- it just means that we have to assess whether what we have achieved is worthwhile. So, maybe we're actually arguing for mostly the same thing: I might be able to interpret your viewpoint as "set the goal to be 100% improvement of all performance metrics". Of course, that goal is impossible, and so whatever we achieve would have to be debated later. My stance is to set an achievable goal, so that we can short-circuit the later discussion. Note that discussing this later is fraught: as the work will already have been done, it is much harder to say "we shouldn't do it".

One thing in common to both approaches, though, is that we can at least set out the benchmarks now. As I said before, I think percentages aren't the most important point -- choosing benchmarks is.

>  some things

What set are these things drawn from? This gets to my desire for benchmarks.

> noticeable amount

What is noticeable? My guess is that you would agree that a 0.001% change is not "noticeable", while a 5% change is. If that's the case, then you have an implicit percentage expectation. I am trying to make that implicit percentage explicit.

> generally doesn't cause regressions

What's a regression? Is a slip of 0.001% a regression? Is a slip of 5% a regression? Again, you probably have an implicit percentage expectation. 

What's "generally"? I interpret this to mean that some regressions somewhere are OK, as long as the regressions aren't everywhere. Maybe this means that you expect the geometric mean performance measurements of xyz to be 0 or better? (And what's "xyz"?)

> Expectations of performance increase in any more detail are more motivations than criteria for success.

I agree here, and thanks for pointing this out. This whole thread started by my asking about the motivation for the proposal. (That's not clear from the history on this ticket, but the comment at the root of this thread was actually internally spun off from https://github.com/haskellfoundation/tech-proposals/pull/1#discussion_r631985001.) While I still think stating a percentage improvement on certain measurements as a goal is fruitful, I'd be content with stating it as a motivation.

I guess, zooming out, that what I'm really after is in making certain implicit assumptions more explicit, lest we all get surprised later when our own internal implicit assumptions differ from others'. The assumptions I'm referring to are things like what measurements to take, what benchmarks to run, what "noticeable" means, what "regression" means, etc. We all have some intuitive sense of this, but let's write it down -- because my guess is that each of us will have slightly different intuitions, and a diversity of unstated assumptions now will cause unhappiness later.

gbaz: > We all have some intuitive sense of this, but let's write it down -- because my guess is that each of us will have slightly different intuitions, and a diversity of unstated assumptions now will cause unhappiness later.

Let me suggest the exact opposite. Suppose there is an overall 20% improvement with no regressions over 1%. Nobody would say "let's not take advantage of that". So the differences in intuitions and assumptions would not matter!

Now if we spent a lot of time now making sure we hashed out all those different intuitions and assumptions, would that change things? No! It would just be a lot of time now arguing about imaginary percentages to get on the same page, and it would turn out those disputes did not matter, because the facts presented a clear picture.

You're right that there are going to be a ton of places where there's room for potential debate after we have some numbers and results in. My view is I think it is _extremely difficult_ to gauge which of those places will be the place debate actually occurs _until_ those results are in. So therefore I would rather not have debates and speculation about things that will not matter, and have the discussions concretely, later, when we know that they are about actual results and not speculative results. I suspect this approach will save a lot of brain cells.

Bodigrim: I've updated sections on Motivation and Performance impact.

proposals/002-text-utf-default.md
```diff
@@ -0,0 +1,307 @@
...
+ends up converting outputs as well (this time UTF-16 to UTF-8).
+
+Even within Haskell ecosystem UTF-16 is rarely used
+for interprocess communication or as a component of binary formats.
+The very `instance Binary Text` serializes `Text` in UTF-8 encoding.
+
+If we switch the internal representation of `Text` from UTF-16 to UTF-8,
+all such conversions would be made redundant and we'll be able just check that
+a `ByteString` is a valid UTF-8 (which is most often the case) and copy it into `Text`.
+If in future (see an upcoming "Unifying vector-like types" proposal) `ByteString`
```
phadej, line 89: I don't like this reference. it sounds as "Unifying vector-like types" is already agreed upon. That is not true.

Bodigrim: I don't feel it says so, it clearly starts with "If". I'd like to put this proposal in a context of other HF developments. I'll improve wording in the next iteration.

proposals/002-text-utf-default.md
neilmayhew: > So my gut feeling is that performance gains will arise mostly from reduced memory pressure.
> That said, for e.g. Russian or Greek text the performance won't be any better. Or any other script which is mostly outside of ASCII (note, not even 256 first codepoints) range.

In the spirit of full disclosure, the scripts used by the majority of languages originating from east of the Mediterranean and now spoken by almost half of the human race require 3 bytes in UTF-8 vs 2 in UTF-16, so the memory pressure will be worse.

_However_, as UTF-8 Everywhere [points out](http://utf8everywhere.org/#asian), real-world documents typically don't increase much in size with UTF-8 vs UTF-16 because they include a lot of ASCII text, in newlines, spaces, punctuation, markup, quotations and so on. This wouldn't apply to shorter pieces of text such as keys, names, words and phrases, but an application would always have the option of using a UTF-16-based text library for that. (A case could even be made for using UTF-16 for `ShortText` since those will have to be created by copying rather than slicing, since the storage model is different, so transcoding won't matter.)

goldfirere: Good points @gbaz, and your argument moved the needle for me. In any case, @Bodigrim's update delivered on what I was hoping for: describing a clear win (> 50% !) for some functions and no regressions elsewhere (defined on a specified testsuite).

Shall we Resolve this conversation then? Given how many have contributed so thoughtfully, I didn't want to just close it out unilaterally. But perhaps the next person to come along, if you agree, just go ahead and close it.

neilmayhew:
> Also, I'd like to have comments why we think UTF8 is better, when e.g.

Those uses of UTF-16 are mostly for historical reasons, because the implementations were done before UTF-8 was common when UCS-2 was still cool (and it was easiest just to upgrade them from UCS-2 to UTF-16 to keep them Unicode-compliant).

proposals/002-text-utf-default.md
goldfirere: > In the spirit of full disclosure, the scripts used by the majority of languages originating from east of the Mediterranean and now spoken by almost half of the human race require 3 bytes in UTF-8 vs 2 in UTF-16, so the memory pressure could be worse in some cases.

Thanks for this information -- I don't know much about character encodings and falsely assumed that the 3-byte UTF-8 slots were for more exotic characters than those used in everyday writing. I apologize.

proposals/002-text-utf-default.md
phadej: I'm worried about the exact content of that proposal. As far as I understood it degraded to something as moderate as "add `primitive` types to `base` and see then what next". I.e. nothing concrete about changing `bytestring` internal representation (it's just changed, not fun changing it again soon, is anyone using `bytestring-0.11` in production?).

proposals/002-text-utf-default.md
```diff
@@ -0,0 +1,307 @@
...
+
+        -   Callan McGill (boarders)
+
+-   Stakeholders:
+
+    -   Edward Kmett: has been vocal about his use of `text-icu` and requires it not be broken.
+
+-   Ben Gamari: integration with GHC
+
+-   The Cabal maintainers (fgaz, emilypi, mikolaj): integration with Cabal
```
phadej, line 147: Neither `Cabal` nor `cabal-install` use `Text` at the moment. It does use (vendored version) of `ShortText` though.

IMO you can just drop `Cabal` from the list, it won't be affected unless there is some need to use `Text` in `Cabal` (which there isn't as far as I can tell).

proposals/002-text-utf-default.md
Bodigrim: We digress, but as you might guess I'm not a fan of changing `bytestring` internal representation as well, `0.11` migration was exhausting. My personal attitude is stated at https://github.com/haskell/bytestring/issues/193#issuecomment-770262177

neilmayhew:
This may be a little premature, but I'd like to make a plea for removing the BOM (Byte-Order Marker) that a lot of Windows software places at the beginning of UTF-8 files. This is used to declare the _encoding_ of the file and isn't part of the _meaning_ of the file. The logical place to do this would be in IO functions such as `Text.readFile` and it could be done efficiently via slicing. If necessary, a non-BOM-eliding version could be provided as well.

As a motivating example, the JSON spec explicitly says that a BOM should not be included in documents, but implementations may choose to skip one if it is present. A surprising number of systems do incorrectly include a BOM (eg Azure). It makes sense to elide it when reading a file or a stream that's already known or required to be UTF-8 because it doesn't contribute anything.

Ban the BOM!

proposals/002-text-utf-default.md
Bodigrim: I don't know much about Cabal internals, but `Cabal.cabal` seems to include `text`:
https://github.com/haskell/cabal/blob/4f8aeb2c8a0a3638e1af887dc869a17e291c8329/Cabal/Cabal.cabal#L272

phadej: It defines few instances for `Text`, it's not really used.

```
cabal master % git grep 'Data.Text'

-- tests:
Cabal-tests/tests/UnitTests/Distribution/Utils/Generic.hs:import qualified Data.Text as T
Cabal-tests/tests/UnitTests/Distribution/Utils/Generic.hs:import qualified Data.Text.Encoding as T

example (note in comment)
Cabal/src/Distribution/Backpack.hs:-- >>> eitherParsec "foo[Str=text-1.2.3:Data.Text.Text]" :: Either String OpenUnitId
Cabal/src/Distribution/Backpack.hs:-- Right (IndefFullUnitId (ComponentId "foo") (fromList [(ModuleName "Str",OpenModule (DefiniteUnitId (DefUnitId {unDefUnitId = UnitId "text-1.2.3"})) (ModuleName "Data.Text.Text"))]))

-- CharParsing class has
-- text :: Text -> m Text
-- method as it's in upstream parsers -package version
Cabal/src/Distribution/Compat/CharParsing.hs:import Data.Text (Text, unpack)

-- these are used for debug, not part of proper build
Cabal/src/Distribution/Fields/Lexer.hs:import qualified Data.Text   as T
Cabal/src/Distribution/Fields/Lexer.hs:import qualified Data.Text.Encoding as T
Cabal/src/Distribution/Fields/Lexer.hs:import qualified Data.Text.Encoding.Error as T

-- ditto
Cabal/src/Distribution/Fields/LexerMonad.hs:import qualified Data.Text          as T
Cabal/src/Distribution/Fields/LexerMonad.hs:import qualified Data.Text.Encoding as T
Cabal/src/Distribution/Fields/Parser.hs:import qualified Data.Text                as T
Cabal/src/Distribution/Fields/Parser.hs:import qualified Data.Text.Encoding       as T
Cabal/src/Distribution/Fields/Parser.hs:import qualified Data.Text.Encoding.Error as T

-- instances
Cabal/src/Distribution/Utils/Structured.hs:import qualified Data.Text                    as T
Cabal/src/Distribution/Utils/Structured.hs:import qualified Data.Text.Lazy               as LT

-- some dev stuff
bootstrap/src/Main.hs:import qualified Data.Text as T
cabal-dev-scripts/src/GenSPDX.hs:import Data.Text        (Text)
cabal-dev-scripts/src/GenSPDX.hs:import qualified Data.Text            as T
cabal-dev-scripts/src/GenSPDXExc.hs:import Data.Text        (Text)
cabal-dev-scripts/src/GenSPDXExc.hs:import qualified Data.Text            as T
cabal-dev-scripts/src/GenUtils.hs:import Data.Text    (Text)
cabal-dev-scripts/src/GenUtils.hs:import qualified Data.Text           as T

-- this is really fun, the usage is
-- let command' = command { commandName = T.unpack . T.replace "v1-" "" . T.pack . commandName $ command }
cabal-install/src/Distribution/Client/CmdLegacy.hs:import qualified Data.Text as T
cabal-testsuite/src/Test/Cabal/Plan.hs:import qualified Data.Text as Text

-- docs
doc/cabal-package.rst:a requirement ``Str`` and an implementation ``Data.Text``, you can
doc/cabal-package.rst:  ``mixins: parametrized requires (Str as Data.Text)``
doc/cabal-package.rst:  ``mixins: text (Data.Text as Str)``
doc/cabal-package.rst:        parametrized (MyModule as MyModule.Text) requires (Str as Data.Text),

-- lexer debugging (same as above)
templates/Lexer.x:import qualified Data.Text   as T
templates/Lexer.x:import qualified Data.Text.Encoding as T
templates/Lexer.x:import qualified Data.Text.Encoding.Error as T
```

If any change affects `Cabal` performance, that will be *interesting*.

phadej: Thinking of `Cabal` made me remember `pandoc`. It can be called *the* Haskell app, and it uses `Text` extensively. So if you want to have some real project on the list, i'd nominate it.

The promise that API doesn't change too much should make building `pandoc` even with its dependency footprint only a matter of CPU time.

Bodigrim: Ah, interesting, thanks.

proposals/002-text-utf-default.md
ketzacoatl: Oh, sorry, I read that as referring to the repo for the package, which would be https://github.com/text-utf8/text-utf8.

proposals/002-text-utf-default.md
emilypi: I agree with @gbaz here. This is more of an implementation detail that I would need @Bodigrim to weigh in on once he can make a reasonable estimate.

proposals/002-text-utf-default.md
emilypi: @goldfirere - There's subtlety in what @davean is saying here: while memory usage may halve in terms of bytes needed to represent text, this does not consider the broader `Text` object (offset + length + array headers), which introduces alignment problems that previously were not there. It's difficult to say whether this will result in an _immediate_ performance increase where say "hey look, here's utf-8!" and suddenly everyone has 2x better memory usage for their text programs. So the distinction is this: 

- We *can* make sure that within the UTF-8 `text` benchmarks we maintain the same fusion guarantees and general asymptotics for its API. 
- We *can't* guarantee it will make Bobby Haskell's immediately 2x faster for all of its text components, because it may fuse or align differently within their program. 
- We *can* guarantee that on the margins, in terms of `text` copying and serialization costs, that memory usage and performance improves. 

So perhaps the distinction needs to be made here in terms of global vs. local benefits. Talking about performance is a minefield unless we set expectations about context. I should reword the point you're citing to reflect that.

emilypi:
@phadej Re: `text-short`, this was my first suggestion for this, having spoken with HVR. However, there's a problem: he's not around, it's not finished, and the library has gone 3 years without a significant update. I suspect that we can have this finished in the same amount of time that it would take us to get `text-short` up to speed and fully implemented.

proposals/002-text-utf-default.md
```diff
@@ -0,0 +1,307 @@
...
+
+We will seek clients' feedback as early as possible, and will act on it, if it arrives
+before the end of the project. However, since our clients are external actors, often
+unpaid volunteers, we cannot expect them to provide feedback by the given date.
+Thus to keep targets of this project time-bound, we cannot include a goal
+of waiting for an approval of indefinite number of parties for indefinitely long.
+Such goal or sentiment, in our opinion, made a significant contribution into failure
+of two previous attempts.
+
+To sum up:
```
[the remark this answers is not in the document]
emilypi: As long as we contextualize what we mean by memory footprint, then yes, I think we can talk about this.

proposals/002-text-utf-default.md
emilypi: That's a very good point. This is an implementation detail that we'll need to solve, and I believe you're hinting at whether the internals should be included in the API.

phadej:
@emilypi the `text-short` into `text` and `text` UTF8ication are not mutually exclusive. In fact, if `Text` is utf-8, for me it makes a lot of sense to integrate `ShortText` into `text` package. The reasoning is the same why there is `ShortByteString`.

I repeat. `Text` is sometimes wrong type, when you don't need slicing. Yes, it adds one more "String" type, but that is the point. The usecases are different. (E.g. if GHC would move to use `text` internally, it would benefit from `ShortText`, not from `Text`).

proposals/002-text-utf-default.md
Bodigrim: I believe that supporting `text-utf16` is up to community. I mean, it's trivial to create such a package: `cabal unpack text-1.2.4.1; cd text-1.2.4.1; mv text.cabal text-utf16.cabal; sed 's/text/text-utf16/g' test-utf16.cabal; cabal sdist; cabal upload`, anyone can do it in a minute. We can put a call for volunteers and bless them to proceed with a fork.

proposals/002-text-utf-default.md
Bodigrim: > Why must `Text` be backed by `ByteArray#`? What are the cons of having a `Text` where decoding from UTF-8 makes no copy?

The main drawback is that pinned memory causes heap fragmentation: a single `ByteString` can easily keep 1Mb block allocated, because GC is not allowed to move it around. 

Indeed it would be nice to back `ByteString` and `Text` by the same structure: either `ByteArray` or `ForeignPtr`, but one way or another this is a far bigger change than UTF-8 transition. It seems that current consensus is that it's up to `ByteString` to change, but anyways it is out of scope of this proposal.

Bodigrim: I've added a section describing API compatibility in more details.

proposals/002-text-utf-default.md
```diff
@@ -118,6 +118,8 @@ to make another attempt.
 
 -   Performance satisfies targets listed below in "Performance impact" section.
 
+-   Compatibiltiy story satisfies targets listed below in "Compatibility issues" section.
```
notquiteamonad, line 121: Typo in Compatibility

```suggestion
-   Compatibility story satisfies targets listed below in "Compatibility issues" section.
```

proposals/002-text-utf-default.md
chrisdone: Aside: whenever I see the fragmentation issue brought up I wonder whether anyone has explored ByteString allocation with jemalloc rather than libc. Its main selling point is low fragmentation. Rust uses jemalloc.

pchiusano:
Very glad to see this happening!!

I really hope the effort doesn't get stymied by endless bikeshedding. 😬

Can anyone explain the relationship with the text-utf8 project? Is the idea to make that a drop-in replacement for the text package, the eventually swap the names around?

Bodigrim:
@pchiusano I'll clarify relations with prior art in an upcoming update. While elaborating the proposal, we came to a conclusion different from current statement.

proposals/002-text-utf-default.md
phadej: @chrisdone `ByteString`s are "malloced" from GHC RTS heap.

chrisdone: Granted, but they consist of a foreign pointer any way; therefore they can take any allocation strategy, and in many cases do when strings come from C via the `unsafePack*` functions.

proposals/002-text-utf-default.md
emilypi: That's a fair split i think @Bodigrim

proposals/002-text-utf-default.md
emilypi: This point is detailed in @Bodigrim's `Compatibility Issues` section.

proposals/002-text-utf-default.md
emilypi: Addressed in recent commits

proposals/002-text-utf-default.md
Bodigrim: As stated in "Motivation" section, at the moment people often use `ByteString` instead of `Text` when looking for ultimate performance. It's easy to convince yourself that you application will never encounter non-ASCII data, if it promises a performance boost. By backing `Text` with UTF-8, we improve performance of ASCII strings and thus assist convergence to using `Text` for anything which is text semantically (and keeping `ByteString` for binary data).

proposals/002-text-utf-default.md
Bodigrim: > If so, is there a reason to make this its own package, and not just e.g. a different datatype exported from the same `text` package?

Keeping both data types in `text` puts a double and infinitely-long burden on `text` maintainers. I don't think it's sustainable, and it's certainly most confusing from users' viewpoint. 

> In reaching out to the community, is there demand for this support? 

As covered in other sections, I suggest we leave this to the community: we are happy to bless and support anyone motivated enough to create a legacy utf16 fork and maintain it as long as desired. But we have neither sustainable resources nor strong motivation to take this burden exclusively ourselves.

proposals/002-text-utf-default.md
Bodigrim: @emilypi shall we remove Cabal from stakeholders?

Bodigrim:
CC @jgm any comments on behalf of `pandoc`?

jgm:
No substantive comments, but I fully support this proposal!

Kleidukos (approved):
Glad to see this taking shape.

proposals/002-text-utf-default.md
```diff
@@ -0,0 +1,392 @@
...
+
+With regards to API compatibility, we intend to keep signatures of non-`Internal`
+modules unchanged, except `Word16` replaced by `Word8` where appropriate.
+Such promise unfortunately cannot be made for `Internal` modules,
+due to their nature: even while we'll strive to keep as much untouched as possible,
+the semantics of internal functions is due to change drastically. This kind of breakage
+should not come as a big surprise, because `Internal` modules have a disclaimer about
+unstable API.
+
+There are two places where `text` leaks details of internal represenation.
```
lorenzo, line 240:
```suggestion
There are two places where `text` leaks details of internal representation.
```

emilypi: Thanks @lorenzo!

phadej:
I should have asked that at the very beginning: What is the goal of having this PR. Is it to facilitate some discussion, which may or may not be incorporated into proposal text based on which (alone, or including the discussion) will be brought to the HF technical group for decision for clearing funds for the implementation?

On the other hand, this project is already on https://haskell.foundation/projects/ page, 
Screenshot from 2021-05-19 20-53-32 with some other projects which don't have a proposal text. So what we are *actually* discussing here? I.e. is it already decided that the project will go forward, and we are discussing some details? What is the process.

emilypi:
@phadej I think that's a meta question that doesn't that shouldn't be answered here, but i'm happy to have an out-of-band discussion about it elsewhere. Please focus on the proposal as a project specification for work that *will* happen if approved of by the HF. The other proposals will operate similarly.

phadej:
The critical point is *if approved*. Thank you. What is the status with the other projects on the https://haskell.foundation/projects/ page, are all of haskup, GHC Platform CI, Project Matchmaker, Performance Tuning Book, Vector Types Proposal and GHC Performance Dashboards also awaiting HF approval (and therefore don't receive funds).

The https://haskell.foundation/projects/ page page gives an impression that all of these projects are in the same state and the impression isn't that it is "awaiting HF approval", but rather that they are already ongoing and "blessed" by HF.

myShoggoth:
Thank you all for the extremely productive feedback on this proposal, it has been great to see people interested and highly involved.

As of the HF Board Meeting on May 20th, 2021 this proposal has been approved.

jberryman:
> If ghc-the-library will ever depend on text directly, re-installing text will be impossible if ghc is not made reinstallable (which I doubt will happen before this change is implemented). So there is that "but" also.

...or indirectly, via `parsec`, which was the case in 9.2:

https://hackage.haskell.org/package/ghc-9.2.1

...but removed by Matt Pickering for us when we hit that wall (https://hackage.haskell.org/package/ghc-9.2.2)

Ericson2314:
GHC is now reinstallable!

Kleidukos:
@Ericson2314 Where / When did this happen?

Ericson2314:
https://gitlab.haskell.org/ghc/ghc/-/merge_requests/5965 GHC *itself*, I suppose I should disclaim :). Stuff like https://gitlab.haskell.org/ghc/ghc/-/merge_requests/6803 is still needed if we are to nicely be able to rebuild all the libraries.

Kleidukos:
thanks!

Bodigrim:
This is old news, but the proposal has been completed successfully in https://github.com/haskell/text/pull/365 and released as `text-2.0`. My ZuriHac talk with experience report is available at https://www.youtube.com/watch?v=1qlGe2qnGZQ.

PR #8, opened 2021-07-06
Add GHC dashboard proposal

emilypi:
Rust's [arewefastyet](https://arewefastyet.com/win10/benchmarks/overview?numDays=60) service is a great example of where we should strive to be, providing users with a sampling of compile-times for a number of commonly used packages over the last several major compiler releases. This serves to "gamify" contributions to compiler performance and provides a means of highlighting the fruits of developers' efforts to users.

While this is a great model, we should note that Haskell's ecosystem poses a few challenges which complicate replicating this approach verbatim:

- lack of stability: while in the case of Rust stability is largely guaranteed across compiler releases, Haskell routinely breaks. This makes it challenging to build the same code across a wide number of compiler releases.
- release schedule: while Rust releases on a six-week cadence, GHC releases on a six-month cycle. This may make release-level reporting too coarse-grained to provide a strong incentive to potential contributors due to the long feedback loop.

Here we propose a set of infrastructure for collecting, recording, and monitoring the compile-time and runtime performance of common Haskell libraries, building on similar infrastructure used by GHC.

[Rendered Proposal](https://github.com/haskellfoundation/tech-proposals/blob/ab4509fa504d776f943205ce706407fa94b07d40/proposals/004-ghc-dashboards.md)

### DIFF
--- a/proposals/004-ghc-dashboards.md
+++ b/proposals/004-ghc-dashboards.md
@@ -1,0 +1,117 @@
+# Are we fast yet? Performance monitoring for the Haskell ecosystem
+
+
+
+## Introduction
+
+Rust's [arewefastyet](https://arewefastyet.rs/) service is a great example of where we should strive to be, providing users with a sampling of compile-times for a number of commonly used packages over the last several major compiler releases. This serves to "gamify" contributions to compiler performance and provides a means of highlighting the fruits of developers' efforts to users.
+
+While this is a great model, we should note that Haskell's ecosystem poses a few challenges which complicate replicating this approach verbatim:
+
+ * lack of stability: while in the case of Rust stability is largely guaranteed across compiler releases, Haskell routinely breaks. This makes it challenging
+   to build the same code across a wide number of compiler releases.
+ * release schedule: while Rust releases on a six-week cadence, GHC releases on a six-month cycle. This may make release-level reporting too coarse-grained to provide a strong incentive to potential contributors due to the long feedback loop.
+
+Here we propose a set of infrastructure for collecting, recording, and monitoring the compile-time and runtime performance of common Haskell
+libraries, building on similar infrastructure used by GHC. 
+
+## Motivation 
+
+Users of Haskell expect that their applications consistently deliver predictable runtime performance across GHC and core library versions. While the
+ecosystem has world-class tools for testing, characterizing and understanding runtime performance (e.g. `Criterion`, `weigh`, `inspection-testing`, to name a few), it has no consolidated infrastructure for tracking performance across the ecosystem over time.
+
+#### Current GHC Infrastructure
+
+This is project proposes to build on top of GHC's current performance tracking infrastructure. The core of this infrastructure is a common schema for
+representing benchmark results and database to record these. This database is fed with samples from a number of sources including:
+
+ * a small set of runtime and compile-time performance testcases in GHC's correctness testsuite, run during the normal continuous integration pipeline
+ * runtime and compile-time metrics from the `nofib` benchmark suite, a subset of which is run during the normal continuous integration pipeline
+ * compile-time statistics collected from builds of the head.hackage patch set, compiled on a nightly basis
+
+Further technical detail on the implementation of this infrastructure will be
+given in a later section.
+
+#### Gipeda
+
+Prior to the introduction of the above-described infrastructure, GHC used the [Gipeda](http://hackage.haskell.org/package/gipeda) package to visualize changes in the metrics from the GHC testsuite collected by a dedicated runner. The `gipeda` package provided a website generator which would produce a static website capable of rendering metric comparisons for pairs of commits. While the interface worked quite well, it was rather limited in temporal range due to the large volume of data that would otherwise need to be generated statically. This limited its utility for long-term tracking of performance characteristics.
+
+## Goals
+
+ The goal of this project is to give GHC authors, core library maintainers, and, eventually, end users insight into their code's performance over time. We hope that this both provides greater understanding of Haskell (especially compile-time) performance characteristics to both GHC developers and users, as well as provides an incentive to spur more contributions in the area of compiler performance.
+
+## People
+
+- **Performers**: Ben Gamari, Matthew Pickering 
+- **Reviewers**: Emily Pillmore 
+- **Stakeholders**: Core libraries maintainers, users
+
+## Resources
+
+There are two resources necessary to make this proposal happen: 
+
+1. The CI cost of machines testing the implementation of the proposal. 
+2. The developer time to implement the infrastructure necessary to achieve the goals of this proposal
+3. The on-going cost of maintaining this infrastructure
+
+While it is expected that cost (2) will likely fall on the shoulders of GHC maintainers, we hope that (3) will eventually be spread across a larger group in due time.
+
+## Timeline
+
+TODO
+
+## Implementation
+
+As mentioned above, the current infrastructure is built upon a shared representation for benchmark results. Specifically, a benchmark sample is a tuple of:
+
+* a git commit hash
+* a string "test environment" characterizing the operating system and hardware on which the test was run
+* a string metric name (e.g. `ghc-testsuite//T1234//bytes_allocated`), which captures, e.g., the name of the test that
+  gave rise to the sample (e.g. a GHC test name like `T1234`) and the name of the collected metric (e.g. `bytes_allocated`)
+* a floating point value. For consistency we generally prefer metrics where growth corresponds to regression.
+
+GHC feeds this database with results produced by continuous integration via a set of GitLab hooks. To expose this database to world we host a read-only [Postgrest](https://postgrest.org/) instance.
+
+#### Proposed Implementation
+
+While the GHC infrastructure described above has been useful for occasional manual analysis, there is a great deal of room for improvement:
+
+1. there is no automated change-point analysis for catching systematic regressions after the fact
+2. no attempt is made to monitor real runtime, instructions, cycles, and other performance characteristics as we lack sufficient capacity of
+   adequately-controlled runner hardware.
+3. the system has no explicit support for hosting measurements from multiple independent, but co-evolving projects (e.g. `bytestring`,
+   which may be affected by changes in itself as well as those in GHC).
+
+To address these, we propose the following concrete technical work:
+
+1. capture metrics from the existing `Cabal` test performed by GHC CI.
+2. collect performance metrics from all GHC commits
+3. provision a readily-available operating system/hardware configuration which can be used for reproducible performance
+   measurement
+4. extend the metric representation to allow for multiple project support, allowing ecosystem-central projects like `text` and
+   `bytestring` to contribute measurements. It is not currently clear how expressive this representation should be.
+5. extend `head.hackage` to allow running and reporting of runtime benchmarks for a subset of widely-used packages
+   (e.g. `text`, `vector`, `aeson`).
+6. implement a scheme (perhaps by enabling write-access in our Postgrest instance) to allow submission of results from outside
+   of GHC's infrastructure.
+
+We expect that most of this work can be done by sufficiently motivated volunteers. The HF's role should principally be organizational.
+
+## Compatibility Issues
+
+N/A
+
+## Performance Impact
+
+N/A
+
+## Deliverables
+
+There are two deliverables associated with this project: 
+
+1. A library, executable, and documentation that will allow the user to stand up performance dashboard monitoring for their project. 
+2. A set of pre-built dashboards for GHC and the core libraries 
+
+## Outcomes
+
+Haskell ecosystem performance will be more observable, leading to more developer eyes on library and compiler performance.


proposals/004-ghc-dashboards.md
```diff
@@ -0,0 +1,117 @@
...
+
+1. there is no automated change-point analysis for catching systematic regressions after the fact
+2. no attempt is made to monitor real runtime, instructions, cycles, and other performance characteristics as we lack sufficient capacity of
+   adequately-controlled runner hardware.
+3. the system has no explicit support for hosting measurements from multiple independent, but co-evolving projects (e.g. `bytestring`,
+   which may be affected by changes in itself as well as those in GHC).
+
+To address these, we propose the following concrete technical work:
+
+1. capture metrics from the existing `Cabal` test performed by GHC CI.
```
goldfirere, line 87: I think I know what this is, but it should be explained.

proposals/004-ghc-dashboards.md
```diff
@@ -0,0 +1,117 @@
...
+2. collect performance metrics from all GHC commits
+3. provision a readily-available operating system/hardware configuration which can be used for reproducible performance
+   measurement
+4. extend the metric representation to allow for multiple project support, allowing ecosystem-central projects like `text` and
+   `bytestring` to contribute measurements. It is not currently clear how expressive this representation should be.
+5. extend `head.hackage` to allow running and reporting of runtime benchmarks for a subset of widely-used packages
```
goldfirere, line 93: Is this redundant with the `Cabal` test?

goldfirere (commented):
I still feel I don't have a good grasp on what is being proposed here. I think the problem is that there is some mixing between specification and implementation. I care much, much more about specification. That is, I would want a proposal of this sort to describe, say, a website where we could monitor GHC's performance (and what features that website would have), as well as a library so that other tools could measure their own performance. A process for how a library can be considered a core part of the ecosystem would also be nice.

As it stands, this all sounds nice, but I am not convinced that accepting and pursuing this proposal would allow me, say, to quickly evaluate how a GHC commit affects compile-time performance in real-world scenarios.

goldfirere:
@doyougnu may be interested.

doyougnu:
Hello, sorry for the novella. 

I am very very interested and would like to help really in any way I can, this
is close to something I've wanted in the Haskell ecosystem for quite some time
so I'm personally invested here.

## Some Ideas on the proposal

### High Level Ideas
- There are really several sub-proposals which relate together different _kinds_
  of deliverables. I see these as the kinds of deliverables:
  1. A front end which displays data
  2. A backend to store data
  3. A collection mechanism to retrieve data

  I think the proposal should separate these out to be more communicative.
  For example:

  1. The front end should have the following features:
    - A public facing UI
    - Plots of an `$entity` by `$metric` by `$condition`
    - where `$metric` can be is `time`, `allocations` etc.
    - and `$condition` can be `commit tag`, `commit`, `date` etc.
    - and an `$entity` is a package
    - Decoupled from all data collection, cleaning and transformation. The front
      end should only plot.

- The proposal really has 3 kinds of users: `ghc-devs`,
  `industrial-haskell-devs`, `everyone-else`. These users value different things
  and are also interested in different metrics. This should be represented in
  the timeline and the deliverables. I propose we define milestones to focus on
  one of these groups at a time. For example:
  - Milestone 0:
    - setup infrastructure _only_ for `ghc-devs`. This means the deliverables
      for this milestone is the full pipeline and CI hooks, but only for `ghc`
      and probably the `core` libraries.
  - Milestone 1:
    - extend infrastructure to miscellaneous packages on Hackage. This allows
      tracking compilation time and other metrics of packages
      `industrial-haskell-devs` care about.
  - Milestone 2:
    - extend infrastructure to an arbitrary package that a miscellaneous
      developer cares about. (this is item 6 of `Proposal Implementation` for
      reference)

### Implementation
Now we can be more specific with the implementation.

#### Front End
Ideally this is going to be a long living project. This means that not only do I
expect the up-time of the front end to be long but I also expect a lot of code
iteration and churn.

##### Requirements
  - Should be in a programming language that most people in the Haskell
    community can pick up, read, and easily understand.
  - Should be easy to deploy, debug, and change. Hot reloading is a major plus.
  - Should reuse existing plotting libraries, and reuse existing web frameworks.
    The front end should be a dead simple web app that simply pulls data from
    the backend and loads the plots.

##### Tech Stack Candidates
  - I think purescript is a great choice here. Haskell-like syntax, good ffi to
    javascript libraries and `D3.js` for visualization.

#### Back End
I'm not familiar with `postgrest` but it seems like a good option for the data
store. From its website it seems like its competitive advantage is the RESTful
api and performance. However, I'm unsure if performance is a major deliverable
for this proposal. Of course, better performance likely means lower CI costs and
more simultaneous users. But depending the architecture of the front end we
might not need a high performance data store. For example, if each plot is a
`csv` file or something similar we could simple store this on disk of the CI
machine and load it as needed. Even for a large amount of data these files can
stay relatively small, for example I just was looking at a `csv` produced by
hacked GHC that had 1.5 million records and was 138Mb.


##### Data Needs
Depending on our data needs we might not even require a database. Checking the
commit history with `git log --all --after="2010-01-01" | grep ^commit | wc -l`
returns ~`41k` commits in the last decade. If we only have one data point per
commit then this isn't that much data. Say we have 20 data points per commit,
then the data would balloon for sure, and it would be expensive to build _once_
but cheap to add to (just append!). Basically I would expect that most data
remains unchanged as time or commits progress.


##### Data Format Requirements
  - Should be a standard format
  - Data must be able to be archived.
  - Should be easily extensible _without_ requiring changes to whatever parses
    and stores the format. To be specific I mean that if we want to parse fields
    `foo`, `bar`, `baz` but there are also extra fields, then the parser should
    still succeed. Similarly, if I want to store an extra field then it should
    be easy to do so without rebuilding or manually editing whatever is storing
    the data.
  - Data should not be stored or parsed by index. The indices are likely to
    change. Consider an example where we have time and alloc plotting setup but
    now we add instructions and cycles, if the indices of time and alloc change
    then suddenly downstream changes are required in the pipeline. Thus the data
    format must be flexible, and the data must be referenced by name.


##### Data Format Candidates
By my requirements `JSON` seems like an obvious candidate. I fear the rigidity
of a database would artificially restrict this project especially in later
milestones. I think it is likely that we'll have a set of collection mechanisms.
For example we might collect allocations and time as reported by ghc with
`-ddump-timings`. But a package owner might want to track a specific function
with `weigh` _just for their package_. Ideally we would be able to support that
and add a _hook_ in the front end that adds a plot for that package. If we use a
database we'll be restricted by the schema, but if we use JSON we can easily
support this extension.

### Purpose of the executable
I'm unsure what your (Emily's) original vision was but I think the executable
should be an easily automated program that calls the collection mechanism(s),
and parses the output into the usable format. This is pretty much the idea
behind my [tool](https://github.com/doyougnu/ghc-build-diagnostics).

The interface should be the shell but with some easily digested commands, but
should not be a shell script in an of itself as this is (mostly) not portable.
Furthermore, an independent developer should be able to manually run the script,
and the user should be able to explicitly pass a path to a custom GHC for the
script to use. This last point serves both the `ghc-devs` and likely the
`industrial-haskell-devs`. Perhaps a good model might be a `Haddock` but instead
of generating documentation it generates the metrics data we would like to track.

### Purpose of the library
I'm unsure what the library should do. I can imagine that it would be a
dependency whose sole purpose is to be used in a benchmark suite. This would
allow a developer to add it separately from the their main project which I think
is crucial. Ideally this library would be similar to `criterion` or `gauge` in
that you can just `Import Foo`, and wrap what you need in `defaultMain` and then
the data we are interested in is produced, saved, uploaded in whatever format we
choose.

Alright, that's all I have, any suggestions appreciated. I can write up a more
formal document and timeline if this sounds reasonable.

goldfirere:
Thanks @doyougnu for the write-up. (Disclaimer: @doyougnu is currently completing an internship at Tweag under my supervision. The subject of the internship is to improve compile times. We did not collaborate on his write-up here.) I really like separating out the classes of users and using e.g. maintenance requirements to inform implementation choices.

However, I'm still not sure how I, as a ghc-dev, will expect to interact with the deliverable. Is it a website? Is it a tool in the GHC source tree? Is it some page at gitlab.haskell.org within its CI output? I recognize that answering these questions may take Real Work that goes beyond what is needed for a proposal to be considered. So maybe there is something smaller that could go in the proposal, something like "At Milestone 0, a GHC contributor will be able to easily see the performance impact of a patch (even on a branch) on major packages in the Haskell ecosystem." Maybe that's not one of the goals -- I'm not well placed enough to know. But whatever the goal is, it must be stated! Then, we can decide how to package the deliverable in a way that will meet the goal.

doyougnu:
@goldfirere this is a good point. If we take `rust` as an example then we'll see that they've separated out `<compiler-devs>`, who work on `rustc` from everyone else. The compiler devs dashboard is hosted on the rust domain at [perf.rust-lang.org](https://perf.rust-lang.org/), the [source](https://github.com/rust-lang/rustc-perf) for that website basically builds a nightly compiler image and then tracks the compile time of the test-suite. This seems pretty similar to the [grafana](https://grafana.gitlab.haskell.org/d/s3PVoM_Mz/ghc-testsuite-performance?orgId=2) dashboard that is already setup. 

To answer your specific question. If you were a `rustc` developer you would follow these instructions:
```
How to compare different versions on your own machine

Often you'll want to compare two different compiler versions. For example, you might have two clones of the rustc repository: one that is unmodified, and a second that contains a branch of your changes. To compare the two versions, do something like this:

./target/release/collector bench_local $RUST_ORIGINAL Original

./target/release/collector bench_local $RUST_MODIFIED Modified

where $RUST_ORIGINAL and $RUST_MODIFIED are paths (relative or absolute) to the relevant rustc executables.
```

which again is pretty similar to what we have now, i.e., having two clones of the compiler, an `original` and a `modified`, then run the benchmarks (or collector in rust world) and compare. To be specific I'm thinking of [these](https://gitlab.haskell.org/ghc/ghc/-/wikis/debugging/ticky-ticky#ticky-ticky-quick-start) instructions in the `ticky` wiki, or `easy.sh` in `nofib`. 

I think the major difference is that there is less friction in the user experience in the rust world. Consider the next section of their wiki:
```
How to view the measurements on your own machine

Once the benchmarks have been run, start the website:

./target/release/site <DATABASE>

wait for the "Loading complete" message to be printed, and then visit localhost:2346/compare.html in a web browser.

Enter the IDs for two runs in the "Commit/Date A" and "Commit/Date B" text boxes and click on "Submit". You can enter the same ID twice, though in that case you won't be shown any percentage differences.

If you've collected new data, you can run curl -X POST localhost:2346/perf/onpush to update the site's view of the data, or just restart the server.
```

Their benchmarking setup provides easy commands to compiler developer to build and run the `perf.rust-lang.org` website locally and observe their local changes _as if_ they had run the CI hooks on `rustc` master. This seems like a good compromise and I like that the local version maintains the commit interface for a comparison.

So in review:
  - the rust community has separated out compiler performance regressions for `rustc` developers and for everyone else. The `rustc` dev's regression seems to be at `perf.rust-lang.org` and for everyone else is at `arewefastyet.rs`
  - The `rustc` dev's tool is made of two executables a `collector` and ` site` the `collector` does the following:
    1. **Input:** two file paths that point to an original and modified version of the rust compiler.
    2. Runs the standard benchmark suite. This is separated into 3 kinds of packages: `Important` packages, `Important and known to stress the compiler` packages, and `Artificial stress test` packages
    3. **Output**: a `sqlite.db` file which is an sqlite database which holds the results of the benchmarks and comparison.
  - The `site` executable does the following:
    1. **Input**: an `sqlite.db` file that stores the benchmarks and comparison
    2. Starts a local web server
    3. Parses the `sqlite.db` file and generates the _same_ plots found at `perf.rust-lang.org` only with the new commit included, hence allowing the comparison.
    4. **Output**: The web server shows the plots at `localhost:<port>//compare.html`

Returning to the proposal, we could use this as a model for milestone 0. In our world the executable would do the exact same thing, the only difference is that we would be locally hosting the `grafana` web site. I think this pipeline makes sense for milestone 0, `ghc-devs` are doing mostly local changes and then running `nofib` or the test-suite anyway and it would be nice to have a visualization based on commits for a locally patched ghc. The use of `sqlite.db` also makes sense since the data produced by these comparison is unlikely to change (and especially because the benchmarks and tests are standardized).

codygman:
One answer I'd like to be able to get or guess at from such data is "Will my 200 route Servant site require 2x the memory usage if I upgrade GHC".

I think that'd be pretty tough to figure out what goes into that and distill it into 'servant-inspired: instance resoution space usage'.

It would be invaluable to preventing issues like [ compile time quadratic in number of routes, with -O2 #986 ](https://github.com/haskell-servant/servant/issues/986#issuecomment-885618332) and narrowing space or time problems to either the library or GHC itself.

chreekat:
o/

While we're here, I thought I'd mention that I am working on improving observability in GHC's pipelines, and my eventual goals will overlap with the goals laid out here.

In fact, I'd say that hiring a DevOps Engineer—another HF proposal that completed successfully (hello! it's me :fairy_man: )—subsumes this project. Rather than trying to design a large system up front, we can set the goals and delegate. Stability and observability of CI are my mission, and if I'm successful we'll fulfill some of the targets of this proposal along the way. I've already added some observability infrastructure for ghc-devs (cf. @doyougnu 's milestone breakdown), in fact.

gbaz:
I also don't know if this discussion addressed the prior art that ghc used to have with gipeda? https://github.com/nomeata/gipeda

and the extension that wasn't deployed for generalizing gipeda outside of ghc:

https://github.com/sgraf812/feed-gipeda

Perhaps dusting those off gives a place to start?

chreekat:
We actually already have residency profiling (of GHC) while compiling every package in head.hackage. There's a demo at https://grafana.gitlab.haskell.org/d/RHebDhq7z/head-hackage-residency-profiling?orgId=2. It needs better performance, better usability, and a lot more visibility. I think that's where I'd start.

david-christiansen:
I'm archiving this proposal because there has not been progress for a long time - it looks like the recent discussion is more "here's something else that's related" rather than "here's progress towards the proposal document". Anyone who is interested in picking it up and continuing should feel free to re-open it!

PR #18, opened 2021-11-29
GHC Ops Proposal

davean:
[Rendered](https://github.com/davean/tech-proposals/blob/main/proposals/000-ghc-ops.md)

In accordance with [the proposal process](https://github.com/haskellfoundation/tech-proposals/blob/be6a8f8bbe4350cd81a13a80cf4f80577c763b81/proposals/PROPOSALS.md) and after circulation on [Discourse](https://discourse.haskell.org/t/pre-hftp-devops-proposal/3747/14), [reddit](https://www.reddit.com/r/haskell/comments/r382bn/prehftp_ghc_devop_proposal/), and [twitter](https://twitter.com/haskellfound/status/1464492635456897024); I submit this proposal to the HFTT.

### DIFF
--- a/proposals/000-ghc-ops.md
+++ b/proposals/000-ghc-ops.md
@@ -1,0 +1,148 @@
+﻿## Introduction
+
+
+This proposal is for hiring a full-time DevOps engineer for GHC and related Haskell projects and tooling. This is necessary to relieve the drain on time of core GHC developers with specialized knowledge on DevOps work. From acceptance to hiring, this proposal should hopefully take two to four months. We see no significant technical risk to this proposal, but one modest concern to be aware of is that transition of DevOps could cause temporary disruption to stability of GHC developer experience.
+
+
+## Background
+
+
+Currently there is a significant amount of time being spent collectively by GHC Operations Contributors keeping systems working. Estimates given in the recent past suggest between Ben Gamari and Moritz Angermann they were spending 20 hours per week on systems.
+
+
+Gitlab takes a moderate amount of attention and issues with it affect a large pool of contributors. Gitlab has had (unplanned) outages on 19 March, and 18 October of 2021. Both were caused by being in a reactive instead of proactive mode since this is no one's job. Full time attention to CI systems can help move into that proactive mode, not only dealing with individual problems as they arise, but stabilizing and making the core machinery more reliable in general -- i.e. dealing with anticipated problems before they occur. (For example, improving garbage-collection of build products and system monitoring).
+
+
+In addition the CI systems are highly varied and numerous. Among them are around 30 runners, across at least 4 kernels - with more user spaces on top. Further we’re missing some CI targets due to a lack of available time and energy from the GHC Operations Contributors to set up and maintain them. GHC CI can not use most hosted CI reasonably because it is already tested on a wider variety of platforms than most provide, and some of its CI jobs far exceed resource limits due to its size.
+
+
+GHC Operations Contributors include but are not limited to:
+* Ben Gamari who is the primary maintainer of our GitLab nix configuration, and both the gitlab and associated gitlab-storage physical boxes. Additionally they run marge-bot and are a primary maintainer of the CI scripts, and several CI boxes.
+* Moritz Angermann handles much of Mac CI, and some CI management in general.
+* davean does some physical Mac cluster management, and supports some of the GitLab servers on a primarily emergency basis.
+
+
+Operational needs extend beyond these core issues. Further basic developer experience tools like the GHC dashboard proposal are stalled for lack of time.
+
+
+## Motivation
+
+
+Hiring a dedicated person to maintain and improve the infrastructure, CI, and related tooling would alleviate the burden on some of GHC's core contributors. This would free these core contributors to make more optimal use of their time instead of spending it on devops - a position which we are more prepared to hire for and requires less onboarding. Further being preemptive in avoiding issues would improve the flow for every contributor. While the freed-up time from contributors might be expended in many productive directions, it is beyond the scope of this proposal to speculate in any detail.
+
+
+Supporting GHC’s process is directly in line with the community glue and adoption directives of the Haskell Foundation -- it connects a wider audience of Haskellers than any other. Additionally it is a challenge for volunteers to reliably sustain this necessary and vital infrastructure. And further, this project should be as far upstream as possible, as GHC development infrastructure is the commons of the Haskell community.
+
+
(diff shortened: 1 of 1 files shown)


goldfirere:
Very excited to see this proposed!

In reading the Discourse thread (thanks @davean for guiding that to a conclusion), there seemed to be an underlying assumption: that hiring a contractor was a short-term solution. I don't think that is necessarily the case.

I see pluses and minuses to the contractor / hire debate. I will write some down:

Pros to hire:
* We have direct control over the hire, and can direct their training, etc., to conform directly to our needs.
* Lower overhead: the contracting org doesn't take a cut.
* Stability: it would seem more likely that the performer would continue in their job for some time. At a contractor, the performer might be reassigned against our wishes.

Pros to contractor:
* Easier to ramp up / ramp down the amount of the contract. For example, if we wanted a blitz of attention early on (say, for 3 months of intensive CI cleaning), we could arrange for that, followed by ongoing maintenance from one person.
* Easier to change skill sets if necessary: if we mis-assess what we'll need at first, we can work with the contracting org to find a different performer who matches our needs better.

The best of both worlds might be to start with a contractor, hone our needs through experience, and then hire once we understand our needs better. Or maybe we just leave this particular decision up to the ED: phrase the proposal to say we wish to have one full-time person paid by the HF -- but the precise nature of the business relationship between the HF and the performer is up to the ED.

The truth is that I don't really know what I'm talking about here, so take all this with a few grains of salt. My big incentive to post is that it seemed, in the Discourse thread, that folks were equating "contractor" with "temporary", and I don't think this equivalence always holds.

Ericson2314:
That's a fair point. I've already been more divisive than I want to but I will say

> Or maybe we just leave this particular decision up to the ED: phrase the proposal to say we wish to have one full-time person paid by the HF -- but the precise nature of the business relationship between the HF and the performer is up to the ED.

Sounds like a good default to me. Unless there is a strong stakeholder preference, e.g. from @bgamari or someone else, I think being open-ended is preferable. Again I happily support this proposal either way, but I like the principle of only codifying the essential details, and leaving the rest to the discretion of the HF staff.

tomjaguarpaw:
Thanks very much for writing this up @davean. It sounds like a great way for the HF to have a huge impact on the whole Haskell community. Hopefully, one benefit for the Haskell community would be that the first releases of major versions of GHC become much more thoroughly tested than they are currently.

Thanks to @rae for elaborating on the contractor vs employee issue. At this stage, I personally don't see the need to determine that the relationship should be one or the other. In fact, an employment relationship brings significant risks to the HF:

* Increasing an organisation's number of employees by 50% is always painful, regardless of the size of the organisation, and can come with significant operational overheads over a period of several months.

* Hopefully everything will go smoothly with the new hire, but if for any reason not then we would suffer some degree of reputational harm if we had to terminate that relationship. It may be legally straightforward to terminate such a relationship (I'm not sure) but it could reflect poorly on the HF regardless of how well the HF conducts itself in the matter.

* If an employee performs poorly enough to be a net drain of resources, but not poorly enough to warrant termination (especially given the risks of termination noted above), then the situation is much harder to resolve than that of a contractor in the same position.

Now, the boundary between employee and contractor is a bit fuzzy. In fact *anyone* hired outside the US may legally *have to* be a contractor. But the distinction I want to draw is between someone that the HF has the moral responsibility to continue the relationship with indefinitely ("employee") and someone it doesn't ("contractor").

N.B. that Ben Gamari is a contractor and I don't get the impression that his work on GHC is in any way worse than it would be if he were an HF employee.  I suggest we open ourselves to the possibility of a similar arrangement in the proposed ops role. 

On that note, I think it would provide additional weight to the proposal if @bgamari and Moritz could add a few words of comment on the current situation and how they envisage things improving with a dedicated ops team member to work with.

gbaz:
I don't think that this proposal specifies the nature of the employment contract, just that this is expected to be an ongoing position, and expected to work a full week. I'm happy to leave the exact nature of the employment in the control of hf officers.

davean:
Do we want to edit the wording in some particular way here, or leave it as a given its up to the HF to execute on this as they see fit?

tomjaguarpaw:
The interpretation that came to mind when I read "Hire a dedicated devops person" and "Deliver one hired developer" was of an "employee" relationship of the nature I described above, i.e. someone to whom the HF has an ongoing moral responsibility. I didn't get the impression that one of the important aspects was "works a full week". Now, that interpretation may be particular to me; others might read it in a different way.

Would it fit what you envisage to express it as "We aim to enter into a long-term engagement with an individual who will work on GHC devops on a full-time basis"?

davean:
Perhaps people find this cleared up now?

bgamari:
For what it's worth, I agree that there isn't a strong reason why the person filling this role would need to be an employee. To the contrary it may be easier if they were a contractor.

As noted in the proposal, managing devops at the scale of GHC is no small task. We target around a dozen platforms, and have a correspondingly large pool of CI runners. Moreover, some of our CI infrastructure could certainly use improvement (e.g. FreeBSD). 

Beyond carrying out maintenance on the runners, it would be nice if the person filling the role would be able to do some amount of work on various pieces of infrastructure surrounding the compiler. For instance, systematically monitoring compiler and runtime performance is a long-standing goal and while progress has been made recently, we still have a longs ways to go. Furthermore, help in diagnosing and triaging flaky tests, improving packaging infrastructure (e.g. putting in place the infrastructure to notarize Darwin builds), and assistance in maintaining head.hackage are all projects which do not require compiler experience but do require consistent, sustained attention. Depending upon the candidate, some or all of these could be areas of contribution.

chreekat:
I think this is great! As somebody fully submersed within the DevOps event horizon, I am certain that the person you hire will have no end of work to do. :) DevOps is a force multiplier that multiplies... itself.

I would suggest that one missing responsibility is "facilitate volunteer DevOps efforts", unless you explicitly want others to be responsible for that.

I also think it would be very cool to include release and deployment to the list of further responsibilities. As written, the proposal seems to be focused on the *testing* piece of the CI pipeline, but DevOps doesn't stop there. Imagine tackling something like the system described at https://github.com/haskell/core-libraries-committee/issues/12#issuecomment-983732316!

I do worry that someone without good project management skills will get bogged down by the sheer size of the problem space.

gbaz:
By unanimous decision of the HFTT committee (meeting of 1 December) we're happy to merge this.

emilypi:
Echoing what @gbaz mentioned - great work @davean! Now it's in @myShoggoth's hands :smile:

PR #21, opened 2021-12-27
Improvements to error messages and their infrastructure

goldfirere:
This proposal sets out an agenda for improving Haskell's error messages. It seeks coordination and consistent energy from an HF person as a way of powering volunteers.

[Rendered](https://github.com/goldfirere/tech-proposals/blob/error-messages/proposals/000-error-messages.md)

(Note: I propose this solely as an individual Haskeller, not in relation to my work with the HF. As that individual Haskeller, I think this is a nicely defined project that would have some real wins. As an HF board member, this looks like a plausibly good use of our time and resources, but there are many other plausibly good uses of our time and resources. I don't claim any particular insight that this is the *best* use of our time and resources, and understand this may reasonably be rejected. I defer to the wisdom of the HFTT members on this point, just as any other proposer would.)

PR #24, opened 2022-01-28
Coordination for structured error messages

goldfirere:
This proposal, a complete rewrite of what was originally #21, seeks coordination from the HF to convert GHC's error messages to use a structured format, benefiting consumers such as HLS. It furthermore seeks to create a website explaining the error messages.

[Rendered](https://github.com/goldfirere/tech-proposals/blob/error-messages-2/proposals/000-error-messages.md)

### DIFF
--- a/proposals/000-error-messages.md
+++ b/proposals/000-error-messages.md
@@ -1,0 +1,220 @@
+## Introduction
+
+This proposal seeks technical coordination from the Haskell Foundation
+for improving the interop story around Haskell tooling error messages.
+While much of the work may be doable by volunteers, the HF would play
+a role in harnessing and corralling the volunteers, as well as coordinating
+common APIs between tools that are both easy to implement and easy to use.
+The HF may also be instrumental in managing an error code namespace, shared
+among all tooling central to Haskell.
+
+## Background
+
+Currently, there is no discipline around error messages. This lack of
+structure manifests itself in a number of ways:
+
+ - Some tools parse the error messages that other tools
+   produce. This is fragile, wasteful, and hard to keep up-to-date. For
+   example, the HLS looks to see if a GHC extension name appears in an error
+   message, in order to allow the user to automatically enable it via a pragma.
+   But since `KindSignatures` is a substring of `StandaloneKindSignatures`, any
+   message mentioning the latter causes HLS to suggest both enabling `KindSignatures`
+   and `StandaloneKindSignatures` -- even though only `StandaloneKindSignatures`
+   would actually work. While there is a workaround here, we can see that
+   better communication between GHC and HLS would avoid this class of problem.
+
+ - Many Haskell error messages refer to advanced concepts. This is unavoidable,
+   as Haskell has advanced features. However, telling a user that their rigid
+   type variable does not unify with a type because there is a kind mismatch
+   is utterly bewildering to Haskell learners. Applying structure to error messages
+   would allow for the creation of an error-message index that could explain
+   what the messages mean -- and how to fix the errors.
+
+ - Given that tools are increasingly working with one another and invoking
+   one another, it can be hard to know who exactly is producing an error
+   message. In one recent example that happened to me, I was trying to get
+   GHC to work with a GHC plugin, and I got a baffling error. It took me
+   more than an hour, if I recall, to discover that the problem was with
+   Haddock (I forget the details) and that I just needed to `--disable-documentation`.
+
+There is already work in this area, within GHC. For the past few years, GHC
+has slowly been converting its error messages to be encoded in data constructors,
+not just as (fancy) strings. [This wiki page](https://gitlab.haskell.org/ghc/ghc/-/wikis/Errors-as-(structured)-values)
+and [this blog post](https://well-typed.com/blog/2021/08/the-new-ghc-diagnostic-infrastructure/) describe
+roughly the state of play. However, this work currently lacks a very important
+ingredient: clients. That is, if GHC is exporting new, fancy datatypes encoding
+its error messages, are these datatypes of use to, say, HLS? We've reached out
+to potential clients for feedback, but the best response we've gotten is something
+along the lines of "sure, looks good". That's encouraging, but I would want to
+a little more coordination to make sure that the interface GHC is building is one
+that can be easily consumed. The HF could help here by coordinating this
+communication between projects.
+
+Furthermore, if this is successful in increasing interop between (say) GHC
+and HLS, then we can expand the idea to other tooling, as well.
+
+The website describing error messages and error-generator identification
+are both fresh in this proposal.
+
+## Motivation
+
+- It is better for tools to collaborate by passing structured data than
+by sending strings back and forth. Structured error messages will thus
+accelerate the development of powerful editor integrations and other
+code analysis tools.
+
+- Establishing a website describing error messages will make it a standard
+reference in the Haskell community and flatten the learning curve to
+new Haskellers.
+
+While the two main goals of this proposal (conversion of all error messages
+to use datatypes; assigning error codes / creating a website) could be
+considered separately, I think they make sense together in this proposal.
+The second goal depends on the first, and it seems likely that many of
+the same potential volunteers will be interested in both. That said, it would
+be fine for, e.g. the HFTT to accept only one part of this proposal without
+the other, or simply not to commit resources until a mid-way review were conducted.
+
+## Goals
+
+1. When compiling a program, HLS queries GHC for error messages and receives
+structured errors, not strings. HLS can then use the information in these structures
+to offer repairs to the user or other options.
+
+1. All GHC error messages include a code. These codes can be searched for on a website
+that explains the error message, with examples of what causes it and how the error
+might be fixed.
+
+1. Stretch goal: Building on the success of the HLS/GHC integration around error
+messages, other central tooling adopts a similar approach. This would, for example,
+enable the possibility that HLS can report more informative configuration errors
+to users, or even to repair some of the problems itself.
+
+1. Stretch goal: The HF would establish a global namespace for Haskell-tool error
+message codes, where each tool includes a code in each message. This would both
+broaden the domain of the website index of error messages and also serve to identify
+the producer of error messages.
+
+## What the Haskell Foundation Would Do
+
+This section is meant to be suggestive of the concrete activity that would support
+this proposal. It is possible the HFTT or other HF people would have an alternative
+approach, which is fine, too.
+
+1. Devote the time of an HF employee (hereby called the Coordinator) to stay on top of
+this project. I think it would be reasonable to timebox this work at 5 hours / week from
+the Coordinator.
+
+1. A key task of the Coordinator is to source volunteers to help with this initiative.
+Accordingly, the Coordinator would be responsible for publicity around this plan, as well
+as thinking creatively about ways to attract volunteers. For example, it might be a fun
+idea to plan a virtual hackathon with potential volunteers or to reward contributions
+with t-shirts. I would expect the Coordinator to think creatively about how to source
+the volunteers. Volunteer management is a primary requirement of the Coordinator; it is
+assumed that the Coordinator is managing volunteers in parallel with all other tasks here.
+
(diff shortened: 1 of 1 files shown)


Ericson2314:
I am very happy to see this! These are indeed the areas I wish to see worked on.

I am skeptical that volunteers will be enough to get the job done, but I don't hold this against the proposal. My gut says that most potential volunteers rather work on end features than these prerequisites which aren't so exciting and of themselves, but gut instincts are not good enough. Either way this goes, I am happy with the outcome:

1. If our experience here proves me wrong, that's *fantastic* --- I would love to see more people come out of the word works interested in these sorts of yak shaves!

2. If we indeed do run into trouble, then we have a good empirical argument that the HF should prioritize doing this deep infrastructural work itself over "last mile" projects which volunteers can tackle.

So big :+1: from me!

goldfirere:
Your point (2) is a very interesting one, @Ericson2314. We've been operating under a general assumption that the HF should primarily coordinate, plan, excite, publicize, etc... and not *do* so much. But maybe this assumption is wrong, and trying (and failing) to simply coordinate on something like this proposal would teach us where we have to *do*.

Of course, the HF may not have the resources to be the primary agent on a project like this, in which case this proposal becomes more aspirational, in the end, than practical. That would be an understandable outcome, and if it fuels a desire e.g. to fundraise in order to support initiatives like this one, that wouldn't necessarily be so bad.

proposals/000-error-messages.md
```diff
@@ -0,0 +1,220 @@
...
+1. All GHC error messages include a code. These codes can be searched for on a website
+that explains the error message, with examples of what causes it and how the error
+might be fixed.
+
+1. Stretch goal: Building on the success of the HLS/GHC integration around error
+messages, other central tooling adopts a similar approach. This would, for example,
+enable the possibility that HLS can report more informative configuration errors
+to users, or even to repair some of the problems itself.
+
+1. Stretch goal: The HF would establish a global namespace for Haskell-tool error
```
michaelpj, line 93: I'm unsure if this is a good idea. We should see if other ecosystems have every managed anything like this. I think _maybe_ Microsoft tried this at some point? There's a risk that this would be quite burdensome for developers and become ignored or obsolete.

proposals/000-error-messages.md
```diff
@@ -0,0 +1,220 @@
...
+
+1. The Coordinator would start by getting an exact handle on the state of structured
+error messages in GHC, by working with current contributors (e.g. Alfredo di Napoli, Sam
+Derbyshire, Richard Eisenberg) and looking at the GHC source code. The Coordinator
+would then identify an area within GHC that would be an appropriate next step to add
+similar structured error messages and source volunteers to contribute to that area.
+
+1. In parallel with the previous item, the Coordinator would work with representatives
+from the HLS team to figure out how HLS might take advantage of the structured error messages
+GHC already has. Even if HLS is not ready to merge yet, the Coordinator and HLS would
```
michaelpj, line 124: I suspect that using the new structured errors will be straightforward. What will probably be non-straightforward and unpleasant is managing the compatibility between versions of GHC that don't have structured errors and those that do. Could be a good GSOC project.

proposals/000-error-messages.md
```diff
@@ -0,0 +1,220 @@
...
+is producing benefits (or is likely to soon).
+
+## Lifecycle:
+
+I don't think this really applies here. There would be a warm-up period at the beginning where the goal
+is to source volunteers, but afterwards, it's all about keeping people moving forwards.
+
+## Deliverables
+
+1. A release of GHC where all of its error messages are structured.
```
michaelpj, line 186: This is happening incrementally, right? Which version of GHC do they start to appear in?

proposals/000-error-messages.md
```diff
@@ -0,0 +1,220 @@
...
+old hands) understand error messages better.
+
+## Risks
+
+- One risk is that the API being built around error messages is not useful to consumers.
+This risk is intended to be mitigated by an early consultation with HLS.
+
+- It is possible that the structured error messages will provide no opportunity for
+improvement over the status quo. This is a risk the HFTT should consider. It might also
+be worthwhile to reach out to HLS now to see what they think.
```
michaelpj, line 212: Oh no, it will be _so_ much better than the _awful_ error message parsing that's going on now. It's buggy, painful, unnecessarily complicated code.

proposals/000-error-messages.md
```diff
@@ -0,0 +1,220 @@
...
+
+- It is possible that no one will find their way to the error-index website, or that
```
michaelpj, line 214: Another reason why the User's Guide is a good location. If we don't think users find it easy to find the User's Guide, then fixing that is great anyway!

michaelpj (approved):
This is great, it would be wonderful to get this effort over the line. My main suggestion is that I think this would be better just scoped to GHC for now. Get GHC's error messages structured and in the User's Guide and that would be a big win on its own.

michaelpj:
> My gut says that most potential volunteers rather work on end features than these prerequisites which aren't so exciting and of themselves, but gut instincts are not good enough.

I think it's worth noting that several of the big PRs to convert GHC errors to structured formats have been from volunteers. I think it's actually quite an appealing thing to work on: you can make a nice difference to the project, and you don't have to know what skolemization is to do it!

Ericson2314:
> > My gut says that most potential volunteers rather work on end features than these prerequisites which aren't so exciting and of themselves, but gut instincts are not good enough.
> 
> I think it's worth noting that several of the big PRs to convert GHC errors to structured formats have been from volunteers. I think it's actually quite an appealing thing to work on: you can make a nice difference to the project, and you don't have to know what skolemization is to do it!

Oh that's great to hear! Thanks for mentioning it.

gbaz:
I mentioned elsewhere that this reminds me of the polymath project: https://polymathprojects.org/

I don't remember where I read it, perhaps Terry Tao's blog, but there's a few good retrospectives about what makes a good polymath project and not, emphasizing the key role of the "coordinator" in managing to divide up worthwhile subproblems that make sense to be tackled by individual volunteers, as well as the need for some sort of "hub" or forum (which in the polymath case is, or at least has sometimes been, just a roundup blogpost with people then jumping in the comments, if i recall). The latter, which i stress can be simple -- anything from a github ticket to a blogpost, to a reddit or discourse post -- is a way to sync everyone up, generate enthusiasm and maintain momentum, and give a place for everyone to share and congratulate one another on their individual contributions. Just some food for thought -- it seemed to make a big difference for polymath, and it may here too.

proposals/000-error-messages.md
```diff
@@ -0,0 +1,220 @@
...
+are both fresh in this proposal.
+
+## Motivation
+
+- It is better for tools to collaborate by passing structured data than
+by sending strings back and forth. Structured error messages will thus
+accelerate the development of powerful editor integrations and other
+code analysis tools.
+
+- Establishing a website describing error messages will make it a standard
```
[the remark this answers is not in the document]
goldfirere: There are pluses and minuses to using the user manual. (I've added a note later in the document that the manual might be a good place.)

Pluses:
 - It seems easier to ensure that the documentation keeps up with the implementation.
 - It seems easier to ensure a high quality of documentation (via the normal PR review).
 - The user manual is already web established, with all of its infrastructure in place.

Minuses:
 - Higher barrier-of-entry to edit the manual, due to GHC's review process.
 - No easy way to just leave a comment. I would see the error-message resource as benefiting from user comments, providing extra examples and insights.
 - Hard to generalize beyond GHC (when that is ready).

proposals/000-error-messages.md
```diff
@@ -0,0 +1,220 @@
+## Introduction
+
+This proposal seeks technical coordination from the Haskell Foundation
+for improving the interop story around Haskell tooling error messages.
```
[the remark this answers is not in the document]
goldfirere: Yes. I started writing with loftier goals, but I realized as I wrote that HLS <--> GHC was the sweet spot. I've toned down the introduction.

proposals/000-error-messages.md
goldfirere: I'd be fine with dropping this from the proposal, but I'm not quite ready to drop it yet, as I really do think that tools should identify the speaker. The "established global namespace" might be something as simple as a two-letter code (or longer) that tools could use if they decide to produce error messages, where the tool decides how to proceed after that prefix.

That said, if others agree that this should be dropped, I'm OK with that.

proposals/000-error-messages.md
goldfirere: Yes, that's a good point. But maybe some design work within GHC will avoid this problem? I'm not sure. And we need to make sure that the errors are delivered in the right way. Maybe part of the reason we didn't get much feedback in the past is that it's obviously going to be easy.

proposals/000-error-messages.md
goldfirere: 9.4 will be the first.

proposals/000-error-messages.md
goldfirere: I agree. But having not heard voices from HLS say this loudly, I didn't want to be over-presumptuous.

Abastro:
Maybe I am polluting this issue, but I don't know where to ask.
Can I browse the involved issues/PRs in the GHC repo? Also is it happening for GHC 9.4 - any validation(backing calculations on ETA) for this?

gbaz:
We met today, but did not have a sufficient quorum to take a decisive vote on this proposal. However, all of us that met were generally in favor of this proposal and think it should pass, and would like to call for an async vote to pass this.

That said, there are some tweaks that we think could be welcome:

* The stretch goal for a "global namespace" is unclear -- is this just adding a tag to each message, identifying which tool it comes from, or is it prefixes to the error codes, or is it a registry between all error codes and tools? Some more detail could be helpful.
* The coordinator is specified to be an "HF employee" but later its specified to be either an employee or another resource (volunteer, in-kind donation, etc). This should be clarified everywhere to the latter.
* An important job of the coordinator will be to shepherd through GHC PRs.
* There's some general concern about who the Coordinator will be -- even if we accept, the HF will need to figure out who to provision and how to get the time commitment.
* There's also some general dubiousness as to if we can actually source the volunteers necessary for this -- however, we think it is a good exercise to try and see how well it can be done!

goldfirere:
Thanks for this feedback, @gbaz. I've updated the proposal accordingly.

proposals/000-error-messages.md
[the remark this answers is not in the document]
cdornan: I would have thought the source of truth would belong in a code repository -- either GHC itself or a dedicated repo. But of course it would be ideal if we could run a tool to extract docs for inclusion in the GHC manaual, which could be incorporated into the GHC manual build pipeline.

Would this not answer to everyone's needs?

cdornan:
I have only just caught sigt of this proposal, and I think it is a great initiative.

I have been a lot of work on structured error messages as part of a wider project that involves looking at the general problem of better structured errors in Haskell. I can't share this work right now, but if folks were interested I could be pushed to speed up plans for publication nd in the meantime provide feedback or help out.

proposals/000-error-messages.md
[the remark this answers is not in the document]
goldfirere: I think we may be agreeing. GHC's manual is in its code repository -- and in a format that can be extracted to the GHC manual website -- so that conforms to your suggestion. Or a separate website would have its content generated from some source repository ("a dedicated repo"), so that also conforms. But I think maybe there is something deeper you're suggesting, but I've missed. Have I captured your intent here?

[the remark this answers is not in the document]
cdornan: I thought as much -- looks like we are all in violent agreement.

gbaz:
Reminder to the committee that we would like an offline vote on this. Committee members only: please reply "yea" or "nea" on this proposal.

kozross:
yea

nomeata:
Yes

Ericson2314:
yea

gbaz:
While we didn't have enough purely offline votes to make quorum, the March 2 meeting of the tech committee (which did have quorum) approved this proposal. The next step, I believe, is to coordinate with the HF to source and appropriate coordinator role to begin implementation.

PR #27, opened 2022-03-07
Introduce GHC.X.hackage proposal

bgamari:
For a few years now the GHC team has maintained [`head.hackage`][head.hackage] a set of patches to a subset of Hackage allowing Hackage packages to be built against pre-release compilers. While `head.hackage` has been an invaluable resource for testing GHC, its user-base has been limited to GHC developers.

In the past few months, there have been a few proposals seeking to lower the cost of migrations to new GHC releases. A common theme in these discussions has been the long lead-time for adoption of new releases due to the largely serial nature of library migration. We propose to extend `head.hackage` as a tool to help users migrate and test their libraries.

[head.hackage]: https://gitlab.haskell.org/ghc/head.hackage/
[Rendered version](https://github.com/bgamari/tech-proposals/blob/ghc-x-hackage/proposals/001-ghc-x-hackage.md)

### DIFF
--- a/proposals/001-ghc-x-hackage.md
+++ b/proposals/001-ghc-x-hackage.md
@@ -1,0 +1,252 @@
+# Adapting to a new version of GHC, way faster
+
+*Michael Peyton Jones, Ben Gamari, Simon Peyton Jones, Théophile “Hécate” Choutri*
+
+
+## 1. Motivation
+
+At the moment we have a problem with propagating support for a new version of GHC through the Haskell ecosystem. Suppose package `A` depends on `B` which depends on `C`. Then the process goes like this
+
+1. GHC version X is released.
+2. The maintainer of `C` must wake up, fix their package, and make a Hackage release. The maintainers of `A` and `B` can do nothing at this point.
+3. The maintainer of `B` must wake up, fix their package, and make a Hackage release. The maintainer of `A` can still do nothing.
+4. The maintainer of `A` can finally fix their package and make a Hackage release.
+
+This process is terrible in lots of ways
+
+- It is utterly serial. The maintainer of `B` cannot lift a finger until the maintainer of `C` has not only fixed package `C`, but also uploaded a new release to Hackage.
+- Each step has multiple serial parts: often the maintainer will merge a patch (perhaps in response to prompting) but _not_ do a release, blocking further progress.
+- If a maintainer is unavailable for any reason, the entire dependency tree of that package is blocked. In an ecosystem with hundreds of widely used packages, the chances of every single maintainter being available in a timely fashion are close to zero.
+- It is not clear to even a willing and available maintainer _when_ they need to wake up and do some work. Often it is up to a motivated individual (say the maintainer of an application that uses `A`) to sequentially bug the maintainers of `C` and `B` and `A` in sequence so they know that they are able to do something.
+
+**The result of this is that it usually takes months or even years for support for a GHC version to percolate through the ecosystem.** For example, GHC 9.0.1 was released in February 2021, but as of December 2021 haskell-language-server still does not support it fully, largely because of difficulty updating dependencies.
+
+We could enormously speed up the process of updating the package ecosystem for a new version of GHC if
+
+- We could parallelise the process.
+- We could allow many people to contribute to the routine updates necessary to adapt a package to changes in its dependencies.
+- We could allow package maintainers to do their work when it was convenient to them, without holding up the whole train
+
+We propose to partially improve the situation by widespread usage of a Hackage overlay (see Section 2), GHC.X.hackage, populated with non-maintainer-contributed patches for the purpose of unblocking package maintainers in adapting their library to new GHC releases.
+
+
+## 2. Background
+
+
+### 2.1 Additional package repositories and Hackage overlays
+
+Cabal supports _additional package repositories_ beyond Hackage. These can be configured in the Cabal config file, or in a cabal.project file.
+
+Users can state which repositories they want to be _active_, and how they should be consulted. This allows additional repositories to override, or merely augment Hackage.
+
+  
+Overriding Hackage can be accomplished by listing the repositories in a Cabal config file, like so:
+```
+active-repositories: hackage.haskell.org, X
+```
+This will cause repository X to be consulted first for any given version of a package, but allow other versions to come from Hackage. For example, if `X` contains `A-1` and `A-2`, and Hackage contains `A-2` and `A-3`, then `A-1` and `A-2` will be taken from `X`, but `A-3` can be taken from Hackage. ([Here is the documentation](https://cabal.readthedocs.io/en/3.6/cabal-project.html) for active-repositories; oddly, the list is searched last-to-first.)
+
+We refer to this style of additional repository as a **Hackage overlay**, since it provides a way to _override_ specific versions of packages which appear in Hackage, while using Hackage for everything else normally.
+
+Additional package repositories behave well in the face of changes. Cabal will recompute the build plan whenever repositories are changed or updated; and since Cabal identifies packages by a hash which includes the hash of the source tarball, Cabal will never get confused between packages from different repositories, and it will cope gracefully even if package sources are _mutated_ in the repository.
+
+
+### 2.2 head.hackage
+
+[head.hackage](https://gitlab.haskell.org/ghc/head.hackage/) is an example of a Hackage overlay. It is built from the linked source repository which consists of a set of patches to released Hackage versions of packages. head.hackage contains patches that enable packages to build with GHC head (hence the name), and is used by the GHC developers to test the upcoming GHC release.
+
+
+## 3. Proposal
+
+
+### 3.1 GHC.X.hackage
+
+
+#### 3.1.1 The GHC.X.hackage package repository
+
+For each released major version _X_ of GHC, a Hackage overlay is provided called GHC.X.hackage. The intention of GHC._X_.hackage is to override versions of packages that require changes to build with GHC-_X_.
+
+GHC._X_.hackage is a _non-authoritative_ set of patched packages, whose goal is primarily to unblock package maintainers. It should not be used in production, and any faults are not to be laid at the door of the package maintainers.
+
+Since GHC._X_.hackage is opt-in, users who do not use GHC._X_.hackage will (obviously) not be able to use the packages which have been fixed in GHC.X.hackage, until fixed versions of those package and all their dependencies are actually released on Hackage.
+
+Additionally, head.hackage becomes a pre-release version of GHC._X_.hackage, and as such will change to behave in the same way as GHC._X_.hackage.
+
+When the release branch for GHC version _X_ is created, the current state of the head.hackage repository is transitioned into GHC._X_.hackage. A new head.hackage is created for the new GHC head.
+
+GHC.X.hackage repositories continue to exist for old versions of GHC in perpetuity. They are not retired, although it is expected that they will be rendered redundant as proper releases are made in Hackage.
+
+
+#### 3.1.2 Adding patches to GHC.X.hackage
+
+The source of GHC.X.hackage is held in a Git repository, and accepts patches from community members _who need not be the package maintainer_. For example, the package maintainer of A might submit patches to fix B and C. This will use the same infrastructure as the existing head.hackage repository: here is an [example of a PR](https://gitlab.haskell.org/ghc/head.hackage/-/merge_requests/194) for head.hackage.
+
+The inclusion criteria for a patch are:
+
+1. The patched package should build with GHC X.
+2. The patch should represent a patch-version level change according to the PVP, in particular it should not change the package’s API.
+3. The patch should be plausibly acceptable by the upstream maintainer.
+
+Criterion (3) is deliberately vague, but suggests that the tests should pass (although this will likely not be easy to verify given the diversity of packages), that the patch should not needlessly break compatibility with previous versions of GHC, etc. Of course, since in most cases the patch will _not_have been vetted by the maintainer, precise correctness cannot be guaranteed.
+
+Package maintainers, or others, can update packages even _before_ GHC X is released, by uploading patches to head.hackage. After the release, those patches will be part of GHC.X.hackage. The same rules for patch inclusion apply to head.hackage before it becomes GHC.X.hackage.
+
+
+#### 3.1.3 Removing patches from GHC.X.hackage
+
+Patches should not _need_ to be removed from GHC.X.hackage. They will simply become obsolete as Hackage becomes more and more up-to-date.
+
+However, in any case if a package maintainer requests the removal of a patch from GHC.X.hackage (for example, it might be outright wrong) then it should certainly be removed.
+
+
+### 3.2 Versioning and releases
+
+While GHC.X.hackage overrides particular versions of packages from Hackage, those overrides will quickly become obsolete when a new version of the package is released to Hackage.
+
+Suppose that GHC.X.hackage contains a patched version of A-X. When A’s maintainer next releases A, they will release A-Y for Y>X. Users who use GHC.X.hackage will in most cases therefore find that cabal picks build plans with the new A-Y instead of the old (patched) A-X. However, if they have constraints that prevent this (say, because A-Y is a major version bump), then they may keep getting the GHC.X.hackage version.
+
+It would be nice to argue that this case is unlikely: since by assumption we only allow patch-level changes into GHC.X.hackage, shouldn’t the next release of A be a patch-level release, and therefore unlikely to be forbidden by the user’s version constraints? But this requires an unrealistic picture of how maintainers work: it is more likely that the maintainer of A has some work-in-progress on A, to which they will _add_ the GHC compatibility patches, to be included in the new release. So in general those changes might reach Hackage in any kind of release!
+
+However, if users are stuck using the GHC.X.hackage versions of packages for longer than strictly necessary that is not too bad: the patches should be inoffensive, and the whole mechanism is opt-in regardless.
+
+
+### 3.3 Package maintainers
+
+Anyone submitting a patch to GHC.X.hackage should also submit a corresponding PR to the source repository of the package. (The PR will usually be identical to the GHC.X.hackage patch.) The maintainer can, _in their own time_, merge the PR and make a release to Hackage.
+
+The only change to the work facing maintainers is that they are hopefully more likely to receive a good-quality patch fixing their compatibility problems for them.
+
+
+### 3.4 Community engagement
+
+The community should be encouraged to engage with GHC.X.hackage both before release (when it is head.hackage) and after release. One could imagine hackathons or other community events, with lists of packages to update, and celebrations when they are all done.
+
+GHC.X.hackage should be announced as part of the GHC X release announcement.
+
+Whenever GHC.X.hackage is mentioned, it must be clearly stated that it is strictly to be used for testing and compatibility work, since it contains patches that have not been vetted by the package maintainers.
+
+
+## 4. Rationale
+
+
+### 4.1 Speed
+
+GHC.X.hackage enables significantly faster and more parallel fixing of packages.
+
+Consider the example from the Motivation section. With GHC.X.hackage, the maintainer of A can, on the day of the GHC X release, fix both B and C in GHC.X.hackage, and use GHC.X.hackage to test and release a fix for A.
+
+However, this speedup _only_ materializes so long as the changes required in A’s dependencies are simple patch-level changes, since those are the only patches that GHC.X.hackage accepts. For everything else, we will still have to await a Hackage release. That may be problematic in practice, since many of the packages which will need major changes to support GHC X are down at the bottom of the dependency tree (e.g. because they make assumptions about GHC internals).
+
+We will have to see in practice how much faster things get: perhaps GHC X will not be usable on day 1, but hopefully it will be usable significantly sooner than it would be otherwise.
+
+
+### 4.2 Community involvement
+
+Because GHC.X.hackage accepts patches from people who are not the package maintainers, it becomes possible for community members to get really involved. A single motivated individual can fix large swathes of the package ecosystem!
+
+If we harness this well, we could potentially enable people to experiment with a mostly-working ecosystem via GHC.X.hackage in a few weeks, rather than months.
+
+
+### 4.3 Maintainer awareness
+
+GHC.X.hackage also partially resolves the problems of maintainers knowing when they are able to do compatibility work. The answer becomes: you can at least _try_ as soon as GHC X is released.
+
+This still isn’t ideal: many packages will be blocked on key blockers that require major changes (and hence Hackage releases), but the current proposal cannot fix that.
+
+
+## 5. Implementation
+
+Since head.hackage already exists, along with all the infrastructure to support it, we believe that it should be relatively straightforward to support multiple GHC.X.hackages, perhaps even with the same source repository (using branches).
+
+There are a couple of roles that need to be filled on an ongoing basis
+
+- Someone needs to maintain GHC.X.hackage. That is: vet patches for inclusion, merge them, keep the CI working, etc.
+- If we want to get significant community involvement, it would be useful for someone to manage that involvement, a “community release manager” of sorts. That is: help people make contributions, drum up excitement, update the community on progress, help upstreaming patches, etc.
+
+There’s a lot of flexibility in filling these roles. They could be
+
+- Occupied by the same person, or different ones
+- Rotating per release, or few releases
+- Filled by a keen community member, or a paid professional
+
+Currently, the GHC.X.hackage maintainer role is effectively filled by the GHC developers who maintain head.hackage. However, if the scope of the role expands, this is likely to be unsustainable. Potentially this is somewhere where the HF could provide assistance.
+
+
+## 6. Alternatives and interactions
+
+
+### 6.1 GHC Maintainer Preview
+
+The[GHC Maintainer Preview proposal](https://github.com/ghc-proposals/ghc-proposals/pull/417) suggested cutting an early pre-release of new GHC versions in order to help maintainers update their packages before the “real” release. A frequent comment in the proposal discussion was that the Maintainer Previews would _not_ actually be directly useful to most maintainers, because they would still be stuck waiting for their (many levels of) dependencies to update before they could do anything. It’s no good having a maintainer pre-release if maintainers can’t _use_ it to fix their packages!
+
+GHC.X.hackage alleviates this problem. Package maintainers could fix their packages and their dependencies by submitting patches to head.hackage, which would become GHC.X.hackage in due course.
+
+
+### 6.2 What’s wrong with head.hackage?
+
+The package upgrade workflow described here is occasionally done today with head.hackage. For example,[here is an attempt](https://github.com/haskell/haskell-language-server/pull/2503) to use head.hackage to pre-emptively fix up haskell-language-server for GHC 9.2.
+
+So maybe we don’t need to do anything except
+
+- Make one head.hackage for each version GHC version _X_
+- Encourage more maintainers to use head.hackage
+- Encourage more people to submit patches to head.hackage
+
+But that’s essentially what the GHC._X_.hackage proposal is, just with a little bit more structure, policy, and marketing.
+
+
+### 6.3 Precognitive releases
+
+The proposal as it stands does not address one of the major roadblocks to getting support out: updating packages that require major changes and hence Hackage releases.
+
+It would be nice to do better than that. We could shoot for the following property:
+
+**Desired Property**. An maintainer of package A can update and release to Hackage a new version of `A`, relying on GHC._X_.hackage for `A`'s dependencies. _This release of `A` should continue to work when the maintainers of those dependencies make their releases to Hackage in due course._
+
+That is, we would like maintainers to be able to make _precognitive_ Hackage releases based on how things _will_ turn out in the future on Hackage.
+
+A sketch of a solution could be:
+
+1. Allow patches in GHC.X.hackage that make more than patch-level changes, and which must create _new_ versions in GHC.X.hackage.
+2. Encourage maintainers to make proactive Hackage releases based on the changes _and versions_ in GHC.X.hackage.
+
+For example, we might have a patch in GHC.X.hackage that fixes `B-2` by creating `B-2.1`. The maintainer of `A` could then release a fixed version of `A` to Hackage, depending on `B < 3.0`, assuming that when `B` is finally updated on Hackage, the compatibility fixes will appear in `B-2.1`.
+
+From this example it is clear that this system relies heavily on _assumptions_ about how package maintainers release new versions of their packages. But this is quite problematic: as discussed in Section 3.2, package maintainers often have work-in-progress which they may want to release _with_ the compatibility patches. So it is quite possible that the next release of B may be B-3, not B-2.1. At that point the maintainer of A has to do _another_ release. That’s terrible - we’ve then made _more_ work for the maintainer of A than they had before GHC.X.hackage existed!
+
+It’s unclear whether a system like this can be made to work. In particular, it is not good enough for the assumptions to be _mostly_ true: even if it _only sometimes_ creates more work for maintainers, this is likely to be enough of a risk to put them off using it.
+
+
+### 6.4 Tested-With
+
+It’s a bad user experience to compile a package with GHC X, and just get a compile error. The `Tested-With` field of a Cabal file helps: if present it specifies which versions of GHC this package has been tested with.
+
+The goal should be that if `A` releases to Hackage a new version of package `A`, depending on (`B < 1.6`), but `B` has not yet released a new version to Hackage, then a naive user who is ignorant of GHC.X.hackage will get a message like
+
+> Package B has not yet been updated to GHC X
+>
+> Consider using GHC.X.hackage (link)
+
+Cabal could produce this message immediately from its solver, without compiling anything.
+
+But perhaps `B` requires no updates at all to work with GHC X (this is a common case). Then this message would be over-conservative. Maybe Hackage could proactively set the `Tested-With` field, by building the package and running its test suite? Or maybe we need two fields: a manual one and an automatic one.
+
+
+### 6.5 Versioning of patched packages
+
+The decision of whether to bump the version number of patched packages is a
+tricky one. While GHC.*X*.hackage patches should avoid making breaking changes
+whenever possible, there will no doubt be times where doing so will be
+unavoidable, even if only in "internal" modules. While the PVP would mandate
+that we mark such changes with a major version bump, doing so would
+significantly hurt the usability of GHC.*X*.hackage for its intended purpose as
+downstream users seeking to test their package against the patch-set would need
+to bump their package's bounds.
+
+In the case of non-breaking changes it is tempting for patches to bump the
+minor version number to reflect the fact that the package is not identical to
+what is found in Hackage. However, doing so brings its own issues. For
+instance, the bumped version may clash with a later version bump made by the
+upstream maintainer. For this reason, we believe that performing no version
+bump at all will lead to the least confusion.
+
--- a/proposals/001-ghc-x-hackage.md
+++ b/proposals/001-ghc-x-hackage.md
@@ -1,7 +1,9 @@
 # Adapting to a new version of GHC, way faster
 
-*Michael Peyton Jones, Ben Gamari, Simon Peyton Jones, Théophile “Hécate” Choutri*
+_Michael Peyton Jones, Ben Gamari, Simon Peyton Jones, Théophile "Hécate" Choutri, David Thrane Christiansen_
 
+
+This proposal describes a way to make it faster and easier to migrate Haskell code with long transitive dependency chains to a new GHC version using an unofficial community repository of compatibility patches.
 
 ## 1. Motivation
 
@@ -12,11 +14,11 @@ At the moment we have a problem with propagating support for a new version of GH
 3. The maintainer of `B` must wake up, fix their package, and make a Hackage release. The maintainer of `A` can still do nothing.
 4. The maintainer of `A` can finally fix their package and make a Hackage release.
 
-This process is terrible in lots of ways:
+This process is slow for a number of reasons:
 
 - It is utterly serial. The maintainer of `B` cannot lift a finger until the maintainer of `C` has not only fixed package `C`, but also uploaded a new release to Hackage.
 - Each step has multiple serial parts: often the maintainer will merge a patch (perhaps in response to prompting) but _not_ do a release, blocking further progress.
-- If a maintainer is unavailable for any reason, the entire dependency tree of that package is blocked. In an ecosystem with hundreds of widely used packages, the chances of every single maintainter being available in a timely fashion are close to zero.
+- If a maintainer is unavailable for any reason, the entire dependency tree of that package is blocked. In an ecosystem with hundreds of widely used packages, the chances of every single maintainer being available in a timely fashion are close to zero.
 - It is not clear to even a willing and available maintainer _when_ they need to wake up and do some work. Often it is up to a motivated individual (say the maintainer of an application that uses `A`) to sequentially bug the maintainers of `C` and `B` and `A` in sequence so they know that they are able to do something.
 
 **The result of this is that it usually takes months or even years for support for a GHC version to percolate through the ecosystem.** For example, GHC 9.0.1 was released in February 2021, but as of December 2021 `haskell-language-server` still does not support it fully, largely because of difficulty updating dependencies.
@@ -27,12 +29,11 @@ We could enormously speed up the process of updating the package ecosystem for a
 - We could allow many people to contribute to the routine updates necessary to adapt a package to changes in its dependencies.
 - We could allow package maintainers to do their work when it was convenient to them, without holding up the whole train
 
-We propose to improve the situation by widespread usage of a Hackage overlay (see Section 2), GHC._X_.hackage, populated with non-maintainer-contributed patches for the purpose of unblocking package maintainers in adapting their libraries to new GHC releases.
-
+We propose to improve the situation by widespread usage of a Hackage overlay (see Section 2), GHC._X_.hackage, populated with **non-maintainer-contributed patches** for the purpose of unblocking package maintainers in adapting their libraries to new GHC releases.
 
 ## 2. Background
 
-### 2.1 Additional package repositories and Hackage overlays
+### 2.1. Additional package repositories and Hackage overlays
 
 Cabal [supports](https://cabal.readthedocs.io/en/latest/cabal-project.html#cfg-field-active-repositories) _additional package repositories_ beyond Hackage. These can be configured in the Cabal config file, or in a `cabal.project` file.
 
@@ -47,11 +48,10 @@ This will cause repository `X` to be consulted first for any given version of a 
 
 We refer to this style of additional repository as a **Hackage overlay**, since it provides a way to _override_ specific versions of packages which appear in Hackage, while using Hackage for everything else.
 
-Additional package repositories behave well in the face of changes. Cabal will
-recompute the build plan whenever new versions are added to any repository.
+Additional package repositories behave well in the face of changes. Cabal will recompute the build plan whenever new versions are added to any repository.
 
 
-### 2.2 head.hackage
+### 2.2. head.hackage
 
 [head.hackage](https://gitlab.haskell.org/ghc/head.hackage/) is an example of a Hackage overlay. It is built from the linked source repository which consists of a set of patches to released Hackage versions of packages. head.hackage contains patches that enable packages to build with GHC `HEAD` (hence the name), and is used by the GHC developers to test upcoming GHC releases.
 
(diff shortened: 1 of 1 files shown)


nomeata:
https://github.com/bgamari/tech-proposals/blob/ghc-x-hackage/proposals/001-ghc-x-hackage.md might be a rendered link

nomeata:
Thanks for writing this up!

The proposal envisions one overlay per major GHC version. But a major GHC version may not be the only case where a large scale migration needs to happen - if a library far down in the dependency tree (or up) changes in ways that need adjustments in many libraries, it seems we are in a similar situation. Could we have a hackage overlays for these as well?

Thinking in the other direction: is having multiple overlays the right call? I see the advantage of not worrying about other GHC versions, and the implicit garbage collection, but there is also redundancy (likely the patch for GHC-9.14 is also needed for GHC-9.16). How would a single “compat patches overlay” repo fare?

tomjaguarpaw:
Sorry to be a petty bureaucrat but [What's the process for submitting a HFTP?](https://github.com/haskellfoundation/tech-proposals/blob/main/proposals/PROPOSALS.md#initial-discussion) says 

> Before submitting a HFTP, it is required that you  ... Discuss your idea on the [Haskell.org Discourse](https://discourse.haskell.org/). ... Create a Discourse topic under the category "Haskell Foundation", that starts with "Pre-HFTP” ... Proposing your ideas on the Discourse is not an optional step 

I don't see such a discussion [on the Haskell Discourse](https://discourse.haskell.org/c/haskell-foundation/11). If I have missed it could you please link it? If not, could you please start the preliminary discussion there?

bgamari:
I have started a discussion [here](https://discourse.haskell.org/t/pre-hftp-ghc-x-hackage-a-tool-for-easing-migrations-to-new-ghc-versions/4239).

TeofilC:
This proposal looks great. I'd be keen to help out once the ball gets rolling.

I have a couple of questions/suggestions:
1. Keeping track of compatible packages
I feel like it would be quite helpful to have a website that shows which packages from hackage build with the help of the overlay. If a package can't be built, it could show the error or a list of failing dependencies.
This would differ from head.hackage's current behaviour as it only tries to build patched packages or those explicitly mentioned, rather than all of hackage.
I think right now it's hard to tell which packages are actually compatible with new GHCs other than by trying to build them yourself. 
And increasing visibility might lead to more patches, and a better idea of ecosystem compatibility.
The release of the new stackage nightly that uses GHC-9.2 seemed to have this effect. It became clearer which packages were failing and led to a flurry of patches. 

2. Stack compatibility
As far as I can tell, the proposal doesn't mention compatibility with `stack`, and I don't think stack supports a similar additional hackage facility like cabal. Do you have plans for stack support?
I think one way this could work is to make available a patched version of the latest lts and/or nightly snapshots.

gbaz:
To update: we met today and we look on this favorably. As it appears active work on fleshing out implementation is being undertaken by the stability working group, we'll hold off on a formal approval until everything shapes up.

Ericson2314:
In the (very long) discussion thread in https://github.com/haskell/core-libraries-committee/issues/22 it came up that doing library impact assessments today is also difficult.

I would like to see `GHC.X.hackage` also help with that, In particular, it should be very easy to make changes `base` on a a branch, make neccesary changes to the corresponding hackage overlay on a branch, and kick off a "rebuild the world" integration test with those two branches.

In particular, a proposed litmus test for pulling the plug and doing the `Data.List` breaking change was first having patches (or merged changed) in all Stackage libraries bringing them into compliance. I think that is a very fine hurdle for breaking changes to clear, and if this proposal makes good infrastructure for that such as I just mentioned, it won't even be a hard hurdle to clear.

Ericson2314:
@Bodigrim left a very nice description in https://github.com/ghc-proposals/ghc-proposals/pull/287#issuecomment-1120173984 how the Simplified Subsumption regression test was botched. That's a bummer, but also makes crystal clear how this proposal can help.

When the way to do such a test is well documented/automated, and the baseline control is well-maintained (so unrelated failures don't ruin the results), doing such experiments will be much easier and more reliable.

chreekat:
Hi! Since I'm the "HF resource" (devops engineer) who might work on this, I thought I'd check on status. I don't yet have a personal opinion one way or the other yet since I'm not really plugged in, but if everybody else has a consensus, maybe we could finish crossing the t's and dotting the i's.

simonpj:
Hi Bryan, thank you.

I suggest:
1. You review the proposal, with Ben, in the light of whatever feedback we have gotten so far -- and of course your own views.
1. Explicitly consult, via direct email, with appropriate folk from
   * stack, 
   * cabal, 
   * hackage
   * core libraries committee

   to check that they are on board.
1. Revise and polish in the light of this feedback.
1. Broadcast a call to library authors to invite their feedback.  "Here's a plan, we'd like to consult you".  In principle they could be contributing now, but there are thousands of library authors, and few of them will be following this repo.
  
  In this phase I'd suggest also writing personally to the maintainers of a few dozen key libraries (i.e. ones with many dependencies).

Actually doing all this takes a bit of bandwidth, which we are always short of, but which your presence will help with a lot.

It's as much about building consensus as about techincal content!

All this is just my suggestion... the rest of the Stability Working Group may have other views.

bgamari:
Indeed, this is a project which we hope to have you work on, @chreekat.

ndmitchell:
I have attempted to use head.hackage in the past, and it didn't work for me. The specific problem I encountered was that a package I depended on had a policy of not releasing to real Hackage, only head.hackage, until a stable GHC was released. That package required a changed API to be compatible with the new GHC, so I was left with the choices:

1) Land the patch in the GitHub repo, and break my ability to make releases if anything else comes up.
2) Don't land the patch in the GitHub repo and make releases to head.hackage instead.
3) Use branches, cherry pick etc. adding lots of cost to my development workflow.

All of those options were unpleasant. The specific package in question was very low down in the tree. I appreciate that head.hackage can parallelise updating dependencies, but uploading to Hackage must be serial. I think it is very important that head.hackage doesn't cause people to delay on uploading to Hackage, which it most definitely has in the past.

chshersh:
As much as I want to reduce the churn around upgrading to a newer GHC, I'm afraid this proposal is a step in the wrong direction. It will waste lots of time by many people, potentially increase the fragmentation of the already fragmented Haskell community, and increase frustration in some areas (even if reducing it in others). I'm going to elaborate in detail on why I think so.

Let's start with flaws in the described workflow:

> 2. The maintainer of C must wake up, fix their package, and make a Hackage release. The maintainers of A and B can do nothing at this point.
> 
> This process is terrible in lots of ways:
> * It is utterly serial. The maintainer of B cannot lift a finger until the maintainer of C has not only fixed package C, but also uploaded a new release to Hackage.

This is not true. Maintainers A and B can submit fixes to package C instead of doing nothing. Both `cabal-install` and `stack` allow depending on specific commits of packages, not only on Hackage versions. So people can contribute patches directly to packages and test those patches in their own packages without waiting for the Hackage release.

> Each step has multiple serial parts: often the maintainer will merge a patch (perhaps in response to prompting) but not do a release, blocking further progress.

We can help maintainers by simplifying Hackage upload workflow or providing necessary CI integrations (e.g. in form of documented CI workflows with examples). This is immediately helpful to **everyone** and will improve the situation straightaway without paying the high cost of this proposal.

> If a maintainer is unavailable for any reason, the entire dependency tree of that package is blocked. In an ecosystem with hundreds of widely used packages, the chances of every single maintainter being available in a timely fashion are close to zero.

The GHC.X.hackage proposal doesn't solve this problem. First of all, this is not a problem in Haskell at the moment since everyone can contribute patches upstream and packages can depend on specific commits. Secondly, if by saying that "the entire dependency tree of that package is blocked" the proposal authors mean "fixes are not on Hackage", the GHC.X.hackage proposal won't help here because it's, well, not Hackage either.

> It is not clear to even a willing and available maintainer when they need to wake up and do some work. Often it is up to a motivated individual (say the maintainer of an application that uses A) to sequentially bug the maintainers of C and B and A in sequence so they know that they are able to do something.

Having a more predictable GHC release schedule helps with that. If maintainers know that they need to wake up e.g. every February 18 to update to a newer GHC, they can sleep in peace for the rest of the year. Another proposal about [tick-tock release cycle for GHC](https://github.com/haskellfoundation/tech-proposals/pull/34) helps here better and moves in the right direction.

<hr>

So to me, it looks like the only feature GHC.X.hackage allows is the ability to write a single line in the package configuration — an alias of a bunch of patches — instead of enumerating all patches one by one. This is a nice benefit but I don't think it's worth the cost.

1. **GHC.X.hackage simply creates double work.** In addition to creating a patch directly to the used package, a contributor also needs to open a patch to GHC.X.hackage (if they're using GHC.X.hackage). Not to mention, that GHC.X.hackage maintainers need to spend time reviewing those patches in addition to the original maintainer reviewing the patch.
2. **GHC.X.hackage widens the fault horizon.** What will happen is that maintainers will always use GHC.X.hackage and we'll effectively have two Hackages. And everyone must look in two places to understand what's going on when things go wrong. And they will go wrong. There's no way to guarantee that the patch proposed to GHC.X.hackage will be the same patch merged to the original package. And when this will happen, it'll create lots of frustrations.

The main explanation of why the Haskell community has delayed GHC support (besides having breaking changes in GHC) is the fact that widely-used packages are maintained by single volunteers. And volunteers don't have to do what you want them to do, they'll do **what they want** and **whenever they want**.

If a maintainer of a widely-used library disappeared from the Haskell community forever, then GHC.X.hackage made its way to perpetuity by having this patch and requiring everyone to always use GHC.X.hackage. Otherwise, GHC.X.hackage just asked lots of people to do lots of redundant work.

<hr>

In conclusion, I feel that this proposal tries to solve the problem of having scarce volunteering resources by requiring volunteers to do even more which obviously doesn't work.

tfausak:
I was in the midst of writing my own comment, but @chshersh said everything I was going to say. (And said it better!) 

For what it's worth, I've never used head.hackage. When upgrading GHC, I've used the Git source feature of both Cabal (`source-repository-package`) and Stack (`extra-deps`) to work around problems.

parsonsmatt:
(I have not read the proposal itself yet, so my thoughts here are entirely untainted with any idea of what we're talking about)

Having just done a bit of work developing compatibility with GHC 9.4 for the `yesod` ecosystem, I actually found the process to be quite smooth. I started with `persistent`, since I maintain that library, and I used `ghcup` to acquire the 9.4 prerelease and begin compiling it. I setup a `cabal.project` and set `--allow-newer`, and traced the build failures. When a package failed to build, I forked it, fixed it, made a PR, and then referred to my fork using `cabal.project`'s `source-repository-package` feature. 

It took about two days of labor to get `yesod`'s tests passing with GHC 9.4.

The primary difficulty, IMO, is updating the strict version upper bounds. I wonder if there's a way to automate that process - ie, if `cabal test --allow-newer` completes successfully with a `base` that is above the bounds, we can bump the bounds. Or if there's a way to "flag" these packages in a semi-automated fashion.

After that, actually implementing the changes needed in `base` and `ghc-prim` were relatively straightforward. The pre-release notes and Migration Guide answered all of my questions.

The main advantage that a `GHC.X.hackage` would have is that it would *share* this work. I have a PR to `yesod` that lists a `cabal.project` that makes it work with GHC 9.4, which you'd copy/paste into your own project, along with depending on that commit of `yesod`, and then you can test your app. But presumably *many* such projects have happened or will happen - not just for `yseod`, but for `servant` or `snap` or `beam` etc. And then every one of those will run into the same build issues with `cereal`, `foundation`, `vector`, etc that need patches. And - unless they see my PRs ot those repos and save themselves the work - may end up duplicating it.

So, if `GHC.X.hackage` allows us to collaboratively provide these patches - then that's great! But if I'm dependent on an absent maintainer (ie `foundation`) to upload to this Hackage overlay, then we're in a bad spot. But why should I (some rando) get to upload to a Hackage overlay, for some package I'm just doing a drive-by contribution for?

Makes me think that the 'proper' fix is to have an index of Haskell packages that are known-to-not-build (either via constraints, `base < 4.17` or via *attempted* builds that fail), linked to a set of potential fixes (that can possibly be community submitted, without guarantee of correctness/completeness - maybe even just links to GitHub PRs or commits). So when I go to build `shakespeare`, it says "The Hackage version is known-to-not-build with GHC 9.4. There's a git commit available here. Want to add this to your `cabal.project`?"

hasufell:
I think there's a plethora of issues:

1. having clear **visibility of upcoming API breakages**, build failures and automatically notifying hackage package maintainers about those (can we detect this automatically? Can this be done semi-automatic as part of GHC release process? How do we notify?)
2. having a central **place for communication**/coordination of migrations. Currently the closest we have to this is stackage issue tracker, e.g. wrt aeson-2: https://github.com/commercialhaskell/stackage/issues/6217
3. an easy way for people to start their work based on the **currently available patchset**

I think the main issue is figuring out build failures and having clear public visibility about those. This is something that should be part of hackage itself. And even point 2 could be argued to be part of hackage.

This proposal *somewhat* fixes point 3, but I think that's not the biggest issue of all. head.hackage could simply just be a `cabal.project.local` pointing to all the upstream PRs instead of another hackage repo. The reason it's a hackage repo is probably because for GHC development it might be unrealistic or unfeasible to create upstream PRs for every single patch. But that is exactly what we want.

parsonsmatt:
OK, reading the proposal and learning what `head.hackage` even is as a 'Hackage Overlay',

> The source of GHC.X.hackage is held in a Git repository, and accepts patches from community members who need not be the package maintainer. For example, the maintainer of package A might submit patches to fix B and C. This will use the same infrastructure as the existing head.hackage repository: here is an [example of a PR](https://gitlab.haskell.org/ghc/head.hackage/-/merge_requests/194) for head.hackage.

The workflow is a bit clunky. I'd really prefer to just say "Here's a GitHub pull request, please make the relevant patch stuff for me." I'm already going to be making a PR upstream with my fixes, might as well share the work here.

One problem with integrating this into my workflow is that many packages I maintain are part of multi-package repositories. So a change to `persistent` means I need to actually, like, figure out how to use `git` in some way and cherry-pick the changes over, for each package that needs to be changed. Implementing a 1:many repo:package relationship support would make this trivially easy: eg, `git clone my sweet persistent fork`, checkout a branch, and generate relevant `patch` files for the relevant packages that need to change.

> The inclusion criteria for a patch are:
> 2. The patch should represent a patch-version level change according to the PVP, in particular it should not change the package’s API.

Hm. Some packages re-export things from `ghc-prim` or `base` in a way that make this impossible. Or, the change to `base` etc require a breaking change or addition to the API. What's the expected solution there?

For the `cabal.project` and `extra-source-repositories` approach, it's fine - I just depend on what I need and fix what needs fixing, making patches as-we-go. A new minor/major version bump as part of upgrading to a new GHC is somewhat expected, IMO.

(ah, this is covered slightly further down in the doc)

> 6.3 Precognitive releases

This whole sections seems really dicey. I think the discussion is mostly correct - the system is too error-prone to be worth any investment of energy. I'm happy having a PR up that links to dependent PRs that require release before a Hackage version can be made, and providing a `cabal.project` or `stack.yaml` that can be used in the interim

> But perhaps B requires no updates at all to work with GHC X (this is a common case). Then this message would be over-conservative. Maybe Hackage could proactively set the Tested-With field, by building the package and running its test suite? Or maybe we need two fields: a manual one and an automatic one.

I've repeated over-and-over again that we have a problem with upper bounds, and that is that we conflate "This is known-not-to-work" and "This is not-known-to-work."

For a bound "This is not-known-to-work," we can always just *test it and see if it works* when the relevant version comes out. Then we either update it to "This is known-to-work" or "This is known-to-not-work."

If Hackage had some support for automagically bumping upper bounds on GHC boot packages (eg `base`, `ghc-prim`, etc) that would be pretty sweet. It seems much more intractable if we're trying to globalize that, though, since you'd need a huge matrix...

The presence of `allow-newer` has made this significantly easier, at least.

# hmmmmm 🤔 

My overall impression is that this is a *lot* more work than my current workflow. It'd be nice to share some of the work (I'm sure I'm not the only one to patch `cereal`, even if I'm the only one to have made a PR), but if folks aren't willing to make PRs upsream with their patches, they're probably *also* not willing to make a PR to a Hackage overlay.

Really, I just want to be able to share `extra-source-repository` stanzas, such that it's trivial for me to contribute one and it's trivial for me to find them.

tfausak:
> Really, I just want to be able to share extra-source-repository stanzas, such that it's trivial for me to contribute one and it's trivial for me to find them.

With Stack, you can create and share a custom snapshot that includes a bunch of `extra-deps` using Git sources. In fact we use this custom snapshot approach to have a unified set of dependencies among internal packages. ([Our snapshot](https://github.com/EdutainmentLIVE/snapshot/blob/f43bf8eb96561e66e2f8cc9940be8e44cc15bd1d/snapshot.yaml#L15) is public, but you probably don't want to depend on it.) The community or the Haskell Foundation or GHC developers could provide a snapshot to rally around when upgrading GHC. 

With Cabal, I don't think it's quite as nice. As @hasufell mentioned, you can cram all the same stuff in a `cabal.project.local`. But I don't think there's a good way to use such a shared config from Cabal directly. People would have to download the file and include it in their project. Not the end of the world, but not as seamless as throwing `resolver: some-url` in your `stack.yaml` file. (Also I think Cabal preemptively downloads all Git sources, whereas Stack only grabs them when needed.) 

For Cabal developers, head.hackage (and this proposal) provide a developer experience that's similar to Stack's. You just add a new repository to your Cabal configuration and you're good to go. That makes me wonder: Is it possible to make an ad hoc Hackage overlay from a `cabal.project` file? That would allow someone to produce a `cabal.project` using `source-repository-package`s for, say, GHC 9.4; and then share that configuration as an overlay. 

I'm just thinking out loud here, so feel free to ignore all this. I mainly wanted to point out that you can already get something like GHC.X.hackage using a custom snapshot with Stack.

hasufell:
> But I don't think there's a good way to use such a shared config from Cabal directly. People would have to download the file and include it in their project.

No, they won't. The next cabal release will support include directives for `cabal.project` including fetching them from remotes.

These may also include resolver constraints, so it's much more ergonomic than using stack for this.

tfausak:
> The next cabal release will support include directives for `cabal.project` including fetching them from remotes.

Great! That was news to me, so I hunted down the pull request: https://github.com/haskell/cabal/pull/7783

brandonchinn178:
> This is not true. Maintainers A and B can submit fixes to package C instead of doing nothing. Both `cabal-install` and `stack` allow depending on specific commits of packages, not only on Hackage versions. So people can contribute patches directly to packages and test those patches in their own packages without waiting for the Hackage release.

I want to echo this. Upgrading my company's codebase to GHC 9 was a not-difficult process, as I can just make forks to upstream repos and reference the new commits in extra-deps.

I'd rather see updates to this issue: https://github.com/haskell/cabal/issues/7821

gbaz:
It sounds like the main concern is duplicate work. Perhaps the process for "constructing" the head overlay can make use of people submitting pointers to patches automatically, so as to reduce duplication here?

hasufell:
> It sounds like the main concern is duplicate work.

I'm not sure it is. As I pointed out I'm not convinced this solves the core issue, which is visibility of required work, overview of current patches and a communication platform.

head.hackage seems to be somewhat specific to GHC workflow.

What makes us think it will help engage less active maintainers? It only talks about the patching workflow.

gbaz:
I want to suggest that a mechanism like this is better than not -- including in terms of an overview of current patches, and visibility of the work.

In particular, if these patches are generated from a git repo, then one gets an overview of all of them and a visibility of them. Further, github tickets on the repo are a reasonable communications platform.

The alternative workflow (what works now) that people are suggesting seems to be that everyone just builds their own list of downstream patches and prs to the things they depend on, ad-hoc. If one collects a centralized collection of such things in a consumable way, then you arrive at, in essence a hackage overlay. So the strict improvement to what exists now seems to lead directly to something like head.hackage (although building it from existing prs when possible rather than forcing duplicate commits of work seems like a very important concern!).

That said, I do find the idea of using a remote cabal.project file rather than a hackage overlay index an interesting technical question. When head.hackage was first developed, we didn't have the tools to do this (including, i believe, git deps) -- and we only fully do now in the upcoming version of cabal, when includes (including remote) in cabal.projects will be available.

Because these features are so new, I really don' t know the ergonomics on either the producer or consumer side of using that approach, but I think it is well worth exploring in terms of this proposal.

hasufell:
> I want to suggest that a mechanism like this is better than not -- including in terms of an overview of current patches, and visibility of the work.

But we're starting the discussion about patch workflow, while there's not even a clear generated list of **broken reverse dependencies**.

This is where e.g. linux distributions start when upgrading toolchain:

1. CI builds reverse deps and collects failures
2. a meta-ticket is posted on the issue tracker about the migration
3. individual tickets to other packages (and possible patches) are added as dependencies of the meta-ticket
4. maintainers are CCed

This proposal doesn't approach the problem in a structured way, imo.

gbaz:
Ah now I see your point -- having a centralized point for patches is nice, but it would be much more useful if accompanied by some centralized forcing function that recorded results of builds of some universe of packages and their blockers. I'm now leaning towards thinking the proposal should discuss how we could get there, even if its somewhat of a stretch goal vs what now exists -- hf proposals can and should be a bit "stretchy" because they're not just about what we can say would be nice with the resources we have, but also what we would aspire to do with more resources -- which in turn can be used by the hf board as a way to reach out and get those resources through corporate donations, etc!

simonpj:
> having a centralized point for patches is nice, but it would be much more useful if accompanied by some centralized forcing function that recorded results of builds of some universe of packages and their blockers

That sounds like a sensible thing to do.  After all, we can hardly engage the community in generating patches that apply simple fixes to packages, if the community does not have a "work-list" to work from.  

But would it not also make sense to have a places to collect such patches, and make them systematically available to others?  That's what this proposal suggests, as I understand it. 

I think that everyone on this thread is looking for constructive suggestions here.  If this proposal isn't going to help, what is?  (Beyond "make no breaking changes, ever", which is a perfectly defensible position, but one that I would deeply regret.)  Specific, concrete suggestions would help us all.

simonpj:
> In conclusion, I feel that this proposal tries to solve the problem of having scarce volunteering resources by requiring volunteers to do even more which obviously doesn't work.

Put like that, it sounds very reasonable.  And perhaps you are right.  But there is another way to look at it.  Perhaps there are lots of volunteers who would be willing to help update packages if only they had a way to help.  Centralising a way for people to work together to generate patch sets across hundreds of packages might unlock that suppressed potential.  And it might make less work for the package authors too, because they just have to merge an already-tested PR, rather than generate it.

I'm not a package author, so I don't know all the dynamics here.   But I'm keen to find ways to relieve the stress on hard-pressed package authors.  Perhaps there are other ways to do that?

hasufell:
> I think that everyone on this thread is looking for constructive suggestions here. If this proposal isn't going to help, what is? 

Well, I think I sketched the required action points broadly in my [previous comment](https://github.com/haskellfoundation/tech-proposals/pull/27#issuecomment-1140490308).

Additionally, the stackage team has been somewhat doing exactly what we're discussing here: building a set of packages in CI, collecting failures, communicating required work and existing patches.

What we would need is simply formalizing and automating this process better, so that both package authors and potential volunteers know what's up and what to do. The details may or may not involve GHC.X.hackage, but I'd argue we need to look at this from a holistic workflow POV instead of pushing isolated solutions.

brandonchinn178:
One additional point I just realized is that not all packages are on GitHub (some packages don't even have a repo link! I had the misfortune of running into such a package recently, and I could only use Hackage's Source links to view the source), so having a centralized platform would make a consistent interface to do so.

But the main point I'm still hung up on is the idea that this proposal would change the "normal" workflow. Right now, on my open source repos, I have a workflow (that I like) where contributors should fork the repo, make a PR, I review it, and merge it. With this proposal, I would also have to somehow be notified when GHC.X.hackage has a patch for my project (there's already an existing issue of maintainers not being contactable via their Hackage info) and copy/paste the patch into my git repo. Then there's two sources of truth for the patch: GHC.X.hackage and the source repo.

To reiterate, IMO, **this proposal solves the wrong problem**. The main bottleneck I experienced is maintainers not merging PRs quickly enough into the canonical repo (indeed, I still have a PR in the monad-validate repo for supporting GHC 9 that hasnt been merged). This proposal doesn't solve that problem; it just provides another mechanism for specifying a patch (which we already have, for GitHub repos). Perhaps fortunately, all the libraries we needed to touch were either on GitHub or GitLab, so maybe that's where my lack of motivation for this proposal is coming from.

Where this proposal really shines (and what it should lay out as the primary motivation) is being a standard interface for providing patches. The current primary motivation specified of things being too serial and whatnot are not issues / already solved (with the caveat of the cabal issue I linked previously). But this proposal should also lay out how it will integrate with maintainers' existing workflows.

One last point I have is looking at other language ecosystems. Rust and Python don't have any notion of an overlay; if packages break on new language versions, you have to fork upstream and patch, which is why you try to choose packages that are actively maintained (which should also be your logic when choosing a Haskell dep). NodeJS doesnt have a notion of an overlay either, but with the recent rollout of Yarn 2, Yarn did provide a way to overlay in the equivalent of stack.yaml/cabal.project ([docs](https://yarnpkg.com/configuration/yarnrc#packageExtensions)), and it also provided a global overlay [here](https://github.com/yarnpkg/berry/blob/master/packages/yarnpkg-extensions/sources/index.ts), so that people didnt have to copy the same overlay in each project. So there is a bit of precedence here, albeit Yarn only dealt with package metadata, not overlaying actual source code.

parsonsmatt:
The mention of Stackage made me wonder: how hard would it be to get something like `Stackage` for GHC HEAD? or even a prerelease stackage? RIght now Stackage Nightly kinda serves that purpose, but it often lags far behind GHC HEAD, often by major versions. A Stackage HEAD could serve this purpose quite nicely.

bergmark:
> The mention of Stackage made me wonder: how hard would it be to get something like `Stackage` for GHC HEAD? or even a prerelease stackage? RIght now Stackage Nightly kinda serves that purpose, but it often lags far behind GHC HEAD, often by major versions. A Stackage HEAD could serve this purpose quite nicely.

I was attempting to create a snapshot for the 9.4 alpha (https://github.com/bergmark/stackage/commit/c0831498d6071b8082ba070c5084897c771d0fea) but I got busy with other things. I think there are a few things missing:

* Metadata updates to [stackage-content](https://github.com/commercialhaskell/stackage-content) (possibly configuring the toolchain to use a branch of it)
* Make sure that [curator](https://github.com/commercialhaskell/curator) can disregard at least upper bounds
* We'll need to apply patches from e.g. head.hackage or git repos. Stackage snapshots themselves work well with git repositories, but I am not sure what needs to be done to get forks/patches into the stackage build pipeline.

gbaz:
> But the main point I'm still hung up on is the idea that this proposal would change the "normal" workflow. Right now, on my open source repos, I have a workflow (that I like) where contributors should fork the repo, make a PR, I review it, and merge it. With this proposal, I would also have to somehow be notified when GHC.X.hackage has a patch for my project (there's already an existing issue of maintainers not being contactable via their Hackage info) and copy/paste the patch into my git repo. 

I don't think this is the case at all, in any version of this proposal. Which is to say that [head.hackage.haskell.org](https://ghc.gitlab.haskell.org/head.hackage/) already exists, and has for a long time! It is expected that it is the responsibility of patch authors to ensure that they also upstream their patches (as it is with debian distro-patches to packages, etc). And the foliage workflow suggested above seems like it would improve that situation, by making, when possible, prs to the existing repo the canonical source for as many ghc.x.hackage patches as possible.

That said, it is well known that the existing system has a clock and gets rolled over and replaced at a certain point, so it does not turn into an ongoing "source of truth".

Where I can see some concerns is that If this proposal tends towards making ghc.x.hackage something used for a "long time" rather than just as an interim measure, people might start to treat it more as an authoritative source rather than a stopgap. I think that such repos should _only_ be used for aiding in ecosystem migrations, and for helping package authors move quickly towards such (and as they are now, for ensuring GHC head has a sufficient universe to stress-test against).

As such, expanding the usage to cover released as well as pre-release ghc seems fine -- but only to a point!  In particular, while we _could_ keep these overlays around "forever" and we currently keep them around insufficiently long, I would  prefer that we still had some "clock" (1 year? 2 year?) on these repos as a forcing function to make sure that people only put out supported released stuff as pinned against central package repos, and not these overlays.

gbaz:
based on our hftt discussion, I think the sentiment was that this proposal doesn't have to extend scope to cover CI or etc, even though those are great additional ideas for further proposals. However, it should be rewritten to focus on the core thing it solves -- sharing work in letting authors update and test packages against newer ghcs (and also giving a gauge as to breakage induced on ghc head), and also to concretely explain the simple workflow for this it enables (building repos out of lists of git hashes or the like).

My concern about "alternate sources of truth" was not widely shared -- rather, people motivated why having these ghc.x.hackage things be long lived was useful for git bisect, ci, etc. That said, I think the proposal still needs to address this as a concern, and motivate why the issue is not that bad (i.e. that the social factor of releasing libraries to hackage as a centralizer will ensure that pinning against odd hackage overlays happens at the leaves of proprietary app code, not on the trunk or branches of the open-source lib ecosystem).

maurobalbi:
@gbaz 
As I understand it, the proposal will allow maintainers to **work** on their packages in parallel, the release to hackage will still be sequential (since packages with overlay-dependencies are not allowed in Hackage)

Ericson2314:
@maurobalbi That is not quite the case. The reasoning is sort of tricky, but if you don't need to bump your bound, you can always push a new version. e.g. if you are working on library `A` and have a `^>= 1.1` dep on library `B`, and the overlay proves a PVP-compatible variant of `B-1.1` is *possible*, you can go ahead and upload your new version of `A`.

The idea is that *if* a new version of `B` is uploaded that is validly within `^>= 1.1`, then your `A` will in fact work with it, and if it supports the newer GHC, and the combination of it and your new `A` will also support the newer GHC.

(The larger story here is that the combination of versioning/what-constitutes-breakage policy and the fact that version bounds include yet-to-be-released packages makes version bounds much more structured than people thing, and gives us theorems like this.)

maurobalbi:
@Ericson2314 Thank you for the clarification. 

> Since GHC.X.hackage is opt-in, users who do not use GHC.X.hackage will not be able to use the packages which have been fixed in GHC.X.hackage, until fixed versions of those package and all their dependencies are actually released on Hackage; however, we hope that by allowing Haskell packages to be made compatible with new compiler releases more quickly, we can help speed this process.

I took this part of the proposal and interpreted it the wrong way.

So in effect the proposal allows maintainers to patch "slow-movers" (without duplicating efforts) and join back into the fold (Hackage)?  That would keep the risk of fragmentation pretty low then. I think that's a nice solution!

Ericson2314:
@maurobalbi well to flip it around, the fragmentation risk is that the new version of `B` never gets uploaded, so while the new `A` doesn't violate the *letter* of the law, it does violate the *spirit* in that people may be using overlays for abandoned packages indefinitely.

I think this OK.

If I recall correctly, we already have some informal ad-hoc procedures for finding new maintainers for abandoned packages. That is good. (I wouldn't want to formalize that process, which is inherently one of tip toeing and special circumstances.) With GHC.X.Hackage, we have an "outlet" for not being blocked before any abandonment becomes acute enough to warrant such a drastic move. And if the time does come to give a package a new maintainer so we can stop using the overlay, there is no better credential for would-be new maintainer than having maintained overlays for that package in the past.

Overall, GHC.X.Hackage is to be about collaborating more, and being blocked less, whereas Hackage has really top-knotch trustees that blur these things but in the original conception is more of a individualist strict package "property rights" regime. I think the combination of all the above will be very fruitful and allow great community bonding with minimal hurt feelings or other drama,

hasufell:
> based on our hftt discussion, I think the sentiment was that this proposal doesn't have to extend scope to cover CI or etc, even though those are great additional ideas for further proposals (and if the authors want to extend scope here somewhat, we'll have no objections!)

How are you going to engage unpaid contributors if they don't even know what is required to do?

If the workflow is not really smooth (and again: submitting patches is not even half of it), then I don't see this taking off beyond GHC core devs and other highly engaged individuals.

Additionally, it'll make it harder for future proposals.

gbaz:
I think the sentiment is that this is a proposal for the _start_ of something, not for the whole of it, and even as is, it is an incremental improvement. But again -- the proposers are very welcome to expand scope, or have some goals as immediate and some sketched as future work. We're just not going to tell them they "must" do any such thing or we'll say no -- at least that's what people are leaning towards. The old adage about the perfect not being the enemy of the good and all that.

david-christiansen:
> Additionally, it'll make it harder for future proposals.

Can you explain a bit more of what you mean here? Which proposals do you see getting more difficult if something like this were implemented? Are there updates to this proposal that could mitigate this risk?

hasufell:
My idea would be to create a proposal which describes the required **process** and interactions, not the required tools.

The tooling can then be an implementation of said proposal (possibly a separate proposal or less formal).

If a proposal with a specific technical solution is accepted, anything that comes after that will be kinda bound/limited to those decisions.

This leads to step by step implementation decisions that lack coherence/vision. 

I don't think anyone here proposes this to be perfect, but things that concern public workflow of unpaid contributors should look at the holistic picture and explain why this will engage contributors.

david-christiansen:
Thanks, @hasufell, your position makes sense to me now. I misunderstood the previous comment quite severely!

david-christiansen:
Here's my summary of the state of the discussion.

# Problem Statement

The lack of backwards compatibility in GHC and `base` releases means that we end up in situations where we would like to use a new compiler version, but cannot because the code that our code depends on is not compatible with the new version.

In particular, some library L might depend on libraries B and C, which themselves depend on library A.

```
     _ B _
    /      \
L -<        >- A
    \      /
     - C -
``` 

In order to make L verifiably work on a new release, A, B, and C must all first be updated, a process that involves their maintainers. These maintainers are typically unpaid volunteers, who will review patches when they please. Most libraries are developed using Git, and most of these are on GitHub. Today, A's maintainer must both update A and perform a release before B and C can proceed, and B and C must both be updated before L can be updated.

Today, many developers work around this by using `cabal-install` or `stack`'s ability to override the package index for specific versions, pointing them at specific commits in which the compatibility problems have been resolved. With this approach, a temporary fork or a pull request can be used as the source of a package, allowing projects further up the dependency chain to start porting before a release has been made. Each developer must figure these overrides out for themselves.

# Proposed Solution

Today, `head.hackage` is a repository of patches against various Haskell packages that allow them to work with the ongoing development version of GHC. It is primarily used by the GHC developers. These patches are used to generate a Hackage overlay, which is a supplementary repository of packages that is used together with Hackage when working with GHC HEAD.

The proposal is to extend this mechanism to add overlays for specific released versions of GHC, as a stop-gap during the porting effort. Instead of each developer needing to configure their build tool to point at specific releases, they could point at GHC.X.Hackage to get patched versions of their dependencies.

# Critiques

## More work for maintainers

Maintainers have objected that this overlay process would increase their workload, rather than decrease it. The fear is that they are expected to both fix the library _and_ send a patch to GHC.X.Hackage. I'm not sure whether this is the intent of the proposal - from what I can see, the intention was that non-maintainers could contribute to GHC.X.Hackage, while maintainers could do a release on Hackage instead. The idea is to provide a way for B's maintainer to update A, and then not require C's maintainer to repeate the process.

## Technical concerns

 * Most developers today do not regularly use patch files - `git format-patch` is not a part of the standard Git curriculum now that sites like GitHub exist. A Git repository full of patch files would introduce more overhead for developers.

 * With Stack, and probably also with the upcoming version of `cabal`, it will be possible to point the tool at an existing list of overrides at some URL. This might be an alternative technical means of attaining the same goal that is more familiar to maintainers' existing workflows. The `foliage` tool seems to have a similar workflow, but it generates Hackage overlays.

 * A Git-centric workflow would rule out packages that aren't in Git. In principle, Hackage doesn't impose a choice of version control system. Supporting _some_ packages may still require arbitrary patch files.
 
## Multiple sources of truth

There is a concern that if GHC.X.Hackage persists long after the release of GHC X, then we end up in a situation where the ecosystem fragments. Some have suggested getting rid of GHC.X.Hackage some fixed amount of time after GHC X is released, but this would also cause tools like `git bisect` to break. I think it would be useful to put in more thought on how to incentivize GHC.X.Hackage being a short-term stop-gap.

## Scope and problem

There has been a critique that this proposal will not create a common work-list of tasks to be done to update the world. The analogy is to a Linux distribution, where updates trigger massive rebuilds with failures reported, so that volunteers have an idea of what works and what doesn't.

Hackage has a few constraints that make this less feasible:
 * Packages on Hackage may require proprietary, internal, or even completely unavailable software to build (e.g. C libraries or network services)
 * Packages may go for a long time without maintenance, but we don't drop them from Hackage, which is append-only
 * Packages on Hackage may depend on incompatible subsets of the rest of Hackage (e.g. one may depend on `L-1.0`, the other on `L-2.0`).

Even if we don't have a central source of knowledge, we do still have a distributed form of knowledge that results from individuals trying to update their code and getting stuck on dependencies. This would also lead to popular libraries being updated more quickly than unpopular ones, which seems useful to me.

It seems to me that this critique is primarily a request for an additional proposal, rather than a reason not to do this particular thing. Is that fair?

------

Is this a fair summary of the discussion so far?

Ericson2314:
Accordingly, this is why I am not really worried about what technical form the initial GHC.X.Hackage takes.

To the extent it is hard to contribute to, regular users will use it without contributing back, but that is fine! Merely having more consumers of the thing will inspire a few people that don't mind the technical hurdles to work on it more anyways --- we can also argue it's firstly GHC team's responsibility to ensure GHC.X.Hackage is good enough so that upon release there is proof to the community that the new GHC is usuable / isn't too disruptive.

If later the GHC team feels overburdened keeping the thing up to date themselves, and the community is clamoring to help out, then we can refine the technical details to make it more convenient, but for the initial step I'm quite happy to just worry about the *social* problem of `head.hackage` only benefiting one group of people, GHC devs, not users.

hasufell:
> Accordingly, this is why I am not really worried about what technical form the initial GHC.X.Hackage takes.

Yes. I'm also leaning towards "just do it" without a proposal. A proper proposal that tries to put the pieces together can happen later.

chshersh:
I personally wouldn't benefit from GHC.X.hackage explicitly (but I might implicitly, only time will show) as I won't be using this overlay by myself. Already existing workflow works well for me. I also like how I can easily see the libraries from my `cabal.project` that still have pending patches. I don't see an easy way to achieve such level of visibility with GHC.X.hackage.

However, this proposal being the HF Tech Proposal, if accepted, may actually make my life more difficult (I can speak only for myself) by:

* Introducing community fragmentation
* Requiring to look in more places when investigating issues
* Having already busy volunteers and maintainers to do more work and making them less productive
* Creating opportunities for people to blame me in poor due diligence if I don't review patches to GHC.X.hackage for my libraries and don't want to engage in this opt-in process
* and much more other _fun_ stuff

I always have this feeling in Haskell that people are eager to dive into complicated tech solutions just to avoid dealing with humans. Yes, dealing with humans is not easy but the entire Haskell ecosystem is built by those humans. It's not sustainable for the entire community if you find it easier to maintain an entire overlay of patches and implement various strategies for retiring patches just to avoid asking maintainers "How could we help you?" or simply understanding that people might disappear.

<hr>

I don't like derailing the conversations around the proposal by suggesting different solutions. But if the ultimate goal is to release new GHC versions faster, then a proposal like the below one helps to achieve the goal in a more community-friendly way:

* https://github.com/haskellfoundation/stability/pull/12

If I'm on my 2-week vacation and someone has a burning need to support new GHC in my library ASAP, I'd rather let them merge their patches directly to my project and release it to Hackage without my involvement at all instead of contributing patches to some external overlay because I'm not available.

<hr>

Addressing some specific details,

> Centralising a way for people to work together to generate patch sets across hundreds of packages might unlock that suppressed potential. And it might make less work for the package authors too, because they just have to merge an already-tested PR, rather than generate it.

I feel like Stackage already does this and it's quite successful. I don't benefit from this process as it requires to use `stack` build tool (which is fair) and I'm using `cabal-install` for my personal projects. And I agree that having a centralised place to track all upstream changes would be a good thing to do and I indeed would benefit from this. However, this feels like a different proposal.

> the current situation is not too bad for a maintainer who:
>
> * Is competent and adept at navigating the Haskell ecosystem (can find broken upstream packages, diagnose them, make and submit fixes, and thinks this is not a big deal)

This can be solved by having better and more visible documentation. Lots of Haskell developers simply don't know they can depend on specific commits of packages even though this feature is available for more than 6 years in the Haskell ecosystem!

> * Does not have too big a dependency footprint (so that fixing their broken upstream packages is feasible)

The assumption here is that the maintainer of a package with lots of dependencies will wake up first and went on a crusade to fix upstream packages. But this is not always the case. I personally wait for at least several months before I even try to upgrade to a newer GHC. Usually, existing GHC works fine for me and I don't have an urgency to upgrade to it.

> * Has the time to do the work (two days of labour for Matt!)

The same is true for GHC.X.hackage. Someone has to contribute a patch. It can be anyone who has free time and enough competency to do the patch. Not necessary the person who needs it. Describing a workflow for contributing patches upstream (e.g. in a form of a blog post but better official guides section) will help lots of people and will significantly decrease the threshold for being an active member of the Haskell community.

> Yes. I'm also leaning towards "just do it" without a proposal

Having HF support for this proposal puts it into completely different perspective. For instance, one of the primary HF goals is to be the glue that connects the entire ecosystem. I don't feel this proposal supports this goal and, in fact, goes against it.

Without the proposal and having official HF support - go for it, you have my blessing 👼🏻 

I'm not in a position to tell volunteers what to do in their free time. If some people want to implement GHC.X.hackage and others want to use it -- good for them 👍🏻 As always, people can do whatever they want in their free time. If it doesn't hurt me and makes life easier for someone else -- I'm happy for them and it doesn't bother me at all 👏🏻

Ericson2314:
@chshersh Well, let's start with the *purpose* of this proposal, which is is simple:

**Foster a way for the community as a whole to collaborate on patched packages so GHC releases are maximally usable on day 1.**

*How* this is accomplished is completely secondary. If the means are too controversial I would advocate splitting the proposal so we can first agree on the goal and then decide on the means. Would that address your concerns?

> I always have this feeling in Haskell that people are eager to dive into complicated tech solutions just to avoid dealing with humans.

Following what I said above, this proposal should instead be at getting us to start dealing with other humans more so we don't just suffer in isolation redundantly creating the same patched package workarounds. It should be *pro* human interaction!

hasufell:
> How this is accomplished is completely secondary. If the means are too controversial I would advocate splitting the proposal so we can first agree on the goal and then decide on the means. Would that address your concerns?

https://github.com/haskellfoundation/tech-proposals/pull/27#issuecomment-1145804124

bgamari:
Thanks to @david-christiansen for summarizing the current state of the discussion. This is extremely useful.

To reiterate a few of his points:

> Maintainers have objected that this overlay process would increase their workload, rather than decrease it. The fear is that they are expected to both fix the library and send a patch to GHC.X.Hackage. I'm not sure whether this is the intent of the proposal - from what I can see, the intention was that non-maintainers could contribute to GHC.X.Hackage, while maintainers could do a release on Hackage instead.

David's interpretation is precisely right. GHC.*X*.Hackage is intended as a means for managing the problem of slow maintainers, which arise for at least two reasons. First, there is time availiability: maintainers have lives, go on vacations, and have busy spells. Such is life and currently we have no good mechanism for managing this fact.

The second reason is a bit trickier: some maintainers are (understandably) reluctant to merge patches updating their packages for a new GHC release until the release is final. Working around this fact was the original motivation for developing `head.hackage`, since otherwise testing GHC prereleases was extraordinarily difficult and required significant duplication of effort.

> A Git-centric workflow would rule out packages that aren't in Git. In principle, Hackage doesn't impose a choice of version control system. Supporting some packages may still require arbitrary patch files.

Indeed this is why we use patch files: they are the least-common-denominator when it comes to representing changes to packages. That being said, I should note that the existing scripts for managing `head.hackage` largely hide patches from the user. Specifically, adding a patch is typically as easy as:
```
$ patch-tool unpack-patch acme-missiles
$ cd packages/acme-missiles
$ # make your change
$ git commit
$ patch-tool update-patches
```


> There has been a critique that this proposal will not create a common work-list of tasks to be done to update the world. The analogy is to a Linux distribution, where updates trigger massive rebuilds with failures reported, so that volunteers have an idea of what works and what doesn't.

It's true that the proposal doesn't attempt to solve the "work-list" problem. However, I would argue that this is an orthogonal problem and one that is to some extent already addressed by the [Hackage matrix builder](https://matrix.hackage.haskell.org/)  (although it seems that it's fallen into quite a state of disrepair, sadly).


@chshersh says,

> If I'm on my 2-week vacation and someone has a burning need to support new GHC in my library ASAP, I'd rather let them merge their patches directly to my project and release it to Hackage without my involvement at all instead of contributing patches to some external overlay because I'm not available.

The Haskell Party Proposal is a nice proposal, but it doesn't address the same problem that we are attempting to address here. It requires that maintainers opt-in, which many maintainers will be understandably reluctant to do. This means that the serial nature of ecosystem evolution is essentially unchanged.

chshersh:
@bgamari 

> The Haskell Party Proposal is a nice proposal, but it doesn't address the same problem that we are attempting to address here. It requires that maintainers opt-in, which many maintainers will be understandably reluctant to do. This means that the serial nature of ecosystem evolution is essentially unchanged.

Apologies for misinterpreting the main goal behind this proposal.

I thought the main goal is to "allow maintainers to upload their libraries to Hackage with the full support of the latest GHC and all dependencies so not-so-advanced Haskell users can use them after the new GHC release as quickly as possible without the need to learn about Hackage overlays" (which, obviously, the GHC.X.Hackage proposal doesn't solve) and not "solve the serial nature of the ecosystem evolution" (which is not solved by the Haskell Party proposal, as you correctly noticed).

In that case, maybe this goal can be highlighted better in the motivation section? E.g. a separate paragraph of bold text after a phrase like "The proposal solves the following problem: ...". Currently, it's too ambiguous in the "Motivation" section and can be misinterpreted as this GitHub thread shows.

hasufell:
> 
> As it stands, the proposal is about de-serialising the workflow. But if there is an even more pressing issue around workflow, maybe someone can sketch out what the problem there is, and how we might address it? To be concrete, here is what I think you may be saying. We need a centralised way to see:
> 
>     * What packages have been patched to work with GHC X
>       
>       * In a Hackage release
>       * As a patch held in GHC.X.Hackage
> 
>     * Which packages need to be patched
> 
>     * Among those that need to be patched, which are most urgent: their dependencies have already been patched, and they have a lot of dependents.
> 
> 
> So a kind of public work-list. Is that what you mean? Or maybe something else. You know far more than I do about the issues here.

Yeah, the way I would envision is it somewhat like this:

## Motivation

This proposal attempts to decrease the adoption time of new GHC releases in the community and make contributing patches to the ecosystem easier.

## Proposal requirements

1. the proposal must primarily decrease the work for maintainers (these are the people in charge of making releases, so that should be the absolute priority)
2. the proposal must show a way on how to connect contributors and maintainers
3. the proposal must describe how these points improve adoption time

## Workflow sketch

1. There must be a publicly visible list of packages that require patching for a specific GHC version
    - the list must include all boot libraries
    - the list should at least consider all packages existing in the latest stackage release
    - the list should expose information about the exact build failures and the build process (cabal flags)
    - the list must be easily searchable
2. There must be an easy way to subscribe to such build failures for a specific package for anyone
     - the subscription must be opt-in
     - there must be a way to easily unsubscribe
3. There must be a central communication platform that is **specific** to fixing build failures of packages wrt new GHC versions
    - there must be a way to search and filter for package related issues on that communication platform
    - this communication platform should be able integrate with a projects main issue tracker (if any)
4. There must be a way to easily submit patches
    - the patch workflow must not require users to submit the same patch in two different places
    - it should be ensured that the maintainer is notified of such patches
5. There must be a way for maintainers to easily merge patches
    - maintainers must be able to understand the context of those changes (e.g. prior discussions, unreleased GHC features, ...)
6. There should be an easy way for maintainers to approve GHC adoption patches ahead of time
    - this process must be opt-in
    - this process must allow to specify constraints (number of reviewers, patch authors, time frame, reverse dep testing, ...)

## Scope

The scope is currently about adopting to newer GHC versions, but could be extended to all ecosystem adoption issues (compare with aeson-2.0 change).

----

My personal opinion is that as much as possible of these solutions should be part of hackage itself.

Ideally, there would be a way to simply upload patches on hackage, then a bot would submit a PR to the maintainers github repository (if any). Hackage would automatically update list of submitted patches, add references to the pull request and somehow monitor changes to the PR itself.

Those things would be nice to have and the underlying tools to achieve the workflow can easily be switched. Hence the proposal focuses on the workflow, not the tools.

bgamari:
> To me it seems that this proposal is more focussed on GHC and less on ecosystem. In that case probably no one disagrees with the proposal, but it should be made clear then that it doesn't attempt to solve the larger ecosystem issues.

I'm not convinced that there is truly much of a difference between these two sets of requirements; I have personally found `head.hackage` to be quite useful while updating my own projects for new GHC releases.

For what it's worth, I agree that your proposal would be useful and would solve another set of problems. However, it strikes me as quite ambitious.

This is of course not a bad thing, but the GHC.X.hackage proposal was intended to be minimal in the sense that it takes something which we already have and which already addresses a clear purpose, and extends it to slightly widen that purpose, allowing more people to benefit from it. The investment needed to reap the benefits from GHC.X.hackage is quite small; however, if the consensus is that these benefits are also quite small then I'm happy to shelve the proposal.

hasufell:
@bgamari apologies... l wasn't really criticizing the proposal on its own. Only the scope/motivation it describes. I think the proposal is an improvement and should be implemented. Versioning the overlays makes sense for various reasons.

My idea would simply be to

1. refine the workflow goals and agree on them
2. prioritize them and figure out which ones can be implemented swiftly to get high return
3. keep the workflow description as an ultimate goal and slowly work towards implementing a coherent user experience

I believe that the GHC.X.hackage proposal can simply be executed in parallel. We might then later agree that it's part of the workflow solution. And even if it's not, it may still serve a purpose for other use cases (such as GHC, bleeding edge CIs etc.).

phadej:
The proposal doesn't seem to mention [Non-maintainer uploads](https://github.com/haskell-infra/hackage-trustees/blob/master/policy.md#3-source-changes-simple-patches). There is clearly an overlap, and it should discuss the differences.

david-christiansen:
I think that the revision to the proposal that @bgamari just pushed at least resolves some of the ambiguities, and hopefully in the process addresses some of the concerns expressed here. In particular, it clarifies that:
 * GHC.X.Hackage will contain non-maintainer-contributed patches 
 * It describes the relationship between the current widespread practice of local version overrides (e.g. in a `stack.yaml` or `cabal.project` file) and GHC.X.Hackage
 * It provides an example workflow to clarify the intended use
 * It specifies that patches should be sent to a package maintainer before including them in GHC.X.Hackage, to minimize divergence over time

Thoughts?

phadej:
The revised proposal still doesn't mention NMUs.

david-christiansen:
@phadej I specifically attempted to address this concern, which makes me think that I have misunderstood you!

I interpret your concern as asking whether patches in the GHC.X.Hackage overlay can be uploaded by non-maintainers. The answer is "yes" - in fact, non-maintainers are expected to be the primary contributors of patches to GHC.X.Hackage. 

But, given your comment, I suspect that you in fact have a different concern related to NMUs. Can you expand upon it a bit? Thanks!

david-christiansen:
Ah, OK, @gbaz explained to me what the concern is - sorry for being a bit dense!

What do you think about the following:
> Hackage has a procedure for non-maintainers uploading package updates in cases where the original maintainer cannot be reached, which is called the _non-maintainer upload_ process (NMU). This process is orthogonal - GHC.X.Hackage is a way for the community to self-organize short-term workarounds to immediate upgrade difficulties, while the NMU process is a way to deal with maintainers who are not reachable at all for a longer period of time. In particular, GHC.X.Hackage can accept patches with a lower degree of scrutiny due to it being a temporary source of workarounds rather than a permanent repository of new versions.

bgamari:
Oh dear, it looks like @david-christiansen and I raced. I just pushed the following language to address NMUs:

> The proposed mechanism is intended to complement Hackage's existing [non-maintainer upload (NMU)
> policy](https://github.com/haskell-infra/hackage-trustees/blob/master/policy.md#3-source-changes-simple-patches),
> where non-maintainers can propose source changes to packages in some cases.
> Specifically, while Hackage NMUs are intended to be carefully-audited
> long-lived releases, GHC._X_.Hackage patches are more transient with a
> correspondingly lower-overhead process and faster turnaround time.

@david-christiansen, do feel free to amend as you see fit.

gbaz:
> GHC.X.hackage repositories continue to exist for old versions of GHC in perpetuity. They are not retired, although it is expected that they will become irrelevant in time as proper releases are made in Hackage.

I would want to add a further element here -- while GHC.X.hackage repos will too exist, I would hope there is a "sunset" (perhaps the release of the following ghc) where they stop accepting packages.

In particular, we wish to encourage that these repos are used strictly for testing and compatibility work -- having a cutoff on accepting new patches would encourage that they not become long-term sources of truth for ongoing non-testing and non-compatibility development.

simonpj:
> I would hope there is a "sunset" (perhaps the release of the following ghc) where they stop accepting packages.

I understand the motivation here but surely the sun can set only when all the packages with patches in GHC.X.Hackage have been updated in Hackage.  And that timetable is in the hands of package authors.  Perhapes it would help if GHC.X.Hackage was accompanied with some kind of status board showing which packages still needed updating in Hackage?  Package authors might find that helpful because they could readily see when their dependencies are all up to date.

If the sun sets (e.g. GHC.X..Hackage is withdrawn for version X) before all packages have been updated then that version X of GHC will abruptly become unavaialble to clients who are depending on those patches.  That would not be a good look.

phadej:
> If the sun sets (e.g. GHC.X..Hackage is withdrawn for version X) before all packages have been updated then that version X of GHC will abruptly become unavaialble to clients who are depending on those patches. That would not be a good look.

I understood what @gbaz said as *stop accepting new patches*, not that the repository itself would go offline.

simonpj:
> I understood what @gbaz said as stop accepting new patches, not that the repository itself would go offline.

Ah, I did not understand that.  If it goes into the proposal, let's make that clear.

But still, it would mean that a user U would be unable to use GHC.X.Hackage to patch an obscure package P, whose maintainer was absent.  Thus GHC version X would be permanently out of reach to U, which again seems undesirable.

gbaz:
> But still, it would mean that a user U would be unable to use GHC.X.Hackage to patch an obscure package P, whose maintainer was absent. Thus GHC version X would be permanently out of reach to U, which again seems undesirable.

Oleg's interpretation is correct. But this would _not_ mean that X would be out of reach to U. They could simply fork U and use a cabal.project or the like to point to the repo. 

It would mean that GHC.X.Hackage would not be a place to continue to perform "testing and compatibility work" for GHC version X. However, this "sunset" would occur only after GHC version X+1 was released, at which point that is where future testing and compatibility would make sense to be directed to. I believe such a sunset would nudge users to treat GHC.X.Hackage as a testing and compatibility area, and not as a permanent source of truth in a way that has relatively minor costs -- i.e. those most affected would only be those who were tempted to use it for unintended purposes to begin with. Further it relieves the GHC.X.Hackage maintainers of the burden of having to accept and review patches for all past GHCs indefinitely.

By the way, I believe the negative concerns expressed in section 6.3 on "precognitive releases" are too negative. Consider if P depends on Q at version 1.1.0 and that fails to work on GHC X, but is patched in GHC.X.Hackage. I claim that releasing a P that depends on _some_ Q 1.1.x (i.e. with pvp upper bounds) remains safe. In the best case, Q 1.1.1 is released promptly, provides compatibility with X, and all is good. If no such Q is ever released, then P's bounds remain correct, and P just remains unbuildable with GHC X. If a Q is later released that is compatible, but it is Q version 2, then if P still builds with it, then only a bounds revision need be performed -- no new release. Only in the case that if Q is released at 2, and P does not work with it, a new version need be released that updates to be compatible with Q -- and the total cost of having made one incremental release is relatively minor, and has done no harm.

bgamari:
It seems that the community is generally somewhat lukewarm with respect to
GHC.X.Hackage. Based on this feedback, and rather than going through the
formal vote process, we're now planning on just doing a little trial to get
some experience: we'll do one iteration of GHC.X.Hackage (perhaps
GHC.9.6.Hackage), see how it works in practice, and evaluate it based on
the reception. If it's a runaway success, great! If there is not widespread
adoption, then we'll be in a position to figure out whether the lack of
adoption is technical (and can be fixed) or if the solution just doesn't
match community needs.

Thank you again for your feedback!

PR #29, opened 2022-04-16
Renumber proposals based on PR number

Ericson2314:
We didn't have a consistent system. ghc-proposals and rust-proposals
previously did something else, but then switched to doing it this way.

~~Draft it case want to add some sort of back-links, too.~~


gbaz:
sounds good to me

PR #32, opened 2022-05-18
Chris has resigned from the group

david-christiansen:


goldfirere (approved):

goldfirere:
I will merge because it's convenient (and fairly obviously correct), while admitting I have no authority here, as I am not in the technical track / working group.

PR #46, opened 2022-11-22
RFC: Ide test integration

santiweight:
This is a proposal for integrating various testing frameworks' output format so that they can be used in IDEs that support Haskell. There is a design for two APIs for test frameworks: `getTestTree` which outlines the test tree of tests in a test suite, and `runTest` which allows for a test group or test case to be run from the test tree.

Feedback is very helpful, especially from testing framework maintainers, so that I can better understand what use cases are most important and we don't miss anything important.

[Rendered](https://github.com/santiweight/tech-proposals/blob/ide-test-integration/proposals/accepted/50-ide-test-integration.md)

### DIFF
--- a/proposals/accepted/50-ide-test-integration.md
+++ b/proposals/accepted/50-ide-test-integration.md
@@ -66,8 +66,7 @@ benefits:
  test trees, and test results.
 2. Generate test trees and test results that can be serialized into JSON from
 Haskell's major testing frameworks, starting with `tasty`.
-3. Integrate test tree output output with HLS (starting with tasty),
-and subsequently VSC, via VSC's [Testing API](https://code.visualstudio.com/api/extension-guides/testing).
+3. Integrate test tree output output with VSC's Haskell extension, via VSC's [Testing API](https://code.visualstudio.com/api/extension-guides/testing).
 4. Stretch goal: support IDEs other than VSC.
 5. Stretch goal: support test results caching, and features such as "Rerun only failed tests".
 
--- a/proposals/accepted/50-ide-test-integration.md
+++ b/proposals/accepted/50-ide-test-integration.md
@@ -1,0 +1,150 @@
+## Abstract
+
+This proposal seeks technical coordination from the Haskell Fondation
+for improving integration between IDEs and testing libraries in Haskell.
+
+This work will involve two major implementation areas: testing library
+integration, and IDE integration. The author will aim to implement both
+of these features, though help is certainly welcome :).
+
+This undertaking requires buy-in and feedback from the maintainers of
+Haskell's testing frameworks, since we would like to have one interface
+that satisfies everyone's needs.
+
+ ## Background
+
+IDE integration for test suites is a feature that is expected in most
+if not all major programming languages. First class support exists for
+[JUnit with multiple IDEs](https://junit.org/junit5/docs/current/user-guide/#running-tests-ide),
+[Python with VSC](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=&cad=rja&uact=8&ved=2ahUKEwiRyt_VwcD7AhUXIUQIHQe1CjUQFnoECBIQAQ&url=https%3A%2F%2Fcode.visualstudio.com%2Fdocs%2Fpython%2Ftesting&usg=AOvVaw0U1x8oEhgUTX1GeoPJqVqo),
+[.NET with VSC](https://marketplace.visualstudio.com/items?itemName=formulahendry.dotnet-test-explorer)
+and many others.
+
+Often these IDEs will supply inline hints for "Run this test" and
+"Re-run failed tests" as well as others. For users of mainstream
+languages, such features are no longer features, but instead
+expectations from a mature language.
+
+Haskell has a mature and well-maintained testing ecosystem, including
+the libraries: `tasty`, `HUnit`, and `hspec`.  While these libraries
+themselves are mature, there is currently minimal
+integration between Haskell's testing libraries and HLS, as well as
+the primary IDEs used for Haskell (VSC;Emacs;Neovim).
+
+Attempts at integration between testing frameworks have focused on
+regex searches, with support for hspec:
+ 1. [neo-test](https://github.com/mrcjkb/neotest-haskell)
+ 2. [vim-test](https://github.com/vim-test/vim-test)
+
+While these are valuable contributions to the ecosystem, their choice
+of regex for the collection of the user's test tree, and support for only
+specific frameworks are symptomatic of a lack of ecosystem coordination.
+
+## Motivation
+
+With the implementation of this proposal, the Haskell community will
+have gained a valuable IDE tool which will improve the user experience
+of Haskellers as well as make Haskellers more productive.
+
+Support for IDE integration of testing libraries will provide the following
+benefits:
+ 1. Ability for IDEs with Haskell support to provide inline hints for "Run
+ Tests" style commands.
+ 2. Ability for IDEs with Haskell support and support for a test explorer
+ view to display the user's test tree and run individual tests.
+ 3. A unified framework for testing libraries to provide integration with
+ IDEs. When a new testing framework is developed, they will be able to
+ provide the results in a Haskell datatype via serialization, which
+ avoids framework-specific test tree discovery logic.
+ 4. Potential to share test result rendering logic between different testing
+ frameworks, whereas currently each testing framework has their own test result
+ display logic.
+
+## Goals
+
+1. Create a new library, `hs-test`, with common datatypes for testing framework
+ test trees, and test results.
+2. Generate test trees and test results that can be serialized into JSON from
+Haskell's major testing frameworks, starting with `tasty`.
+3. Integrate test tree output output with HLS (starting with tasty),
+and subsequently VSC, via VSC's [Testing API](https://code.visualstudio.com/api/extension-guides/testing).
+4. Stretch goal: support IDEs other than VSC.
+5. Stretch goal: support test results caching, and features such as "Rerun only failed tests".
+
+## Technical Proposal
+
+The `hs-test` library will contain the following data declarations:
+
+```haskell
+-- | Datatype representing all the tests that exist in a test
+-- suite that could be run.
+data TestTree =
+  TestGroup {name :: Text, subTrees :: [TestTree]}
+  | TestCase {name :: Text}
+
+data RunTestsRequest =
+  RunTestsRequest {path :: NonEmpty Text}
+
+-- | A lazy list in order to support arbitrary execution
+-- order and streaming between the test runner and IDE
+data TestResults = TestResults [TestCaseResult]
+
+data TestCaseResult = TestCaseResult {
+    -- | Represents the path to get to this test case from the
+    -- TestTree, where each element represents a node in the original
+    -- TestTree.
+    path :: NonEmpty Text,
+    result :: TestResult,
+    timeTaken :: Maybe Double
+  }
+
+data TestResult = Success | Skipped | Failure FailureReason
+
+newtype FailureReason = FailureReason Text
+```
+
+It will also contain the logic to encode this datatype
+to JSON, for consumption by IDEs.
+
+------------
+
+For testing frameworks that want to support IDE integration,
+they should provide support for the following commands:
+
+```bash
+> cabal test test-suite-name \
+      \--test-options='test-tree --output="FILE_NAME"'
+> cabal test test-suite-name \
+      \--test-options='run-tests --input="FILE" --output="FILE"'
+```
+
+The `test-tree` command outputs a JSON-serialized `TestTree` to the given
+`output` file.
+
+The `run-tests` command reads a JSON-serialized `RunTestsRequest` from
+`input`, and, after executing tests, writes the result to the `output` file.
+
+`hs-test` will provide helper functions for these serialization tasks.
+
+-------------
+
+For IDEs, they will utilize these commands to find the test tree, execute
+the test tree, and display the results to the user.
+
+## Future Work
+
+This proposal is purposefully minimal in scope, for a number
+of reasons:
+ 1. In order for testing frameworks to buy in, we will need a
+ minimally-controversial interface.
+ 2. Support for more features should be done incrementally, since
+ the design space here is rather large, and we will not know users'
+ needs until after integration is alright out in the real world.
+
+Here are some features that have been neglected in this proposal:
+ 1. Distinction between "error" and "failure" when a test fails.
+ 2. Support for multiple filters when running a test.
+ 3. Support for options when running a test, such as: `--accept` when
+ running golden tests, `-jN` to run tests in parallel, and options
+ related to property-based tests.
+ 4. Support for localized options when running a test.
--- a/proposals/accepted/50-ide-test-integration.md
+++ b/proposals/accepted/50-ide-test-integration.md
@@ -78,9 +78,12 @@ The `hs-test` library will contain the following data declarations:
 ```haskell
 -- | Datatype representing all the tests that exist in a test
 -- suite that could be run.
+--
+-- `SrcLocs` are used to indicate where inline hints should be displayed
+-- in an IDE.
 data TestTree =
-  TestGroup {name :: Text, subTrees :: [TestTree]}
-  | TestCase {name :: Text}
+  TestGroup {name :: Text, subTrees :: [TestTree], location :: SrcLoc}
+  | TestCase {name :: Text, location :: SrcLoc}
 
 data RunTestsRequest =
   RunTestsRequest {path :: NonEmpty Text}
@@ -112,10 +115,14 @@ For testing frameworks that want to support IDE integration,
 they should provide support for the following commands:
 
 ```bash
+./test-suite test-tree --output="FILE_NAME"
+./test-suite run-tests --input="FILE" --output="FILE"
+
+# Sample usages:
 > cabal test test-suite-name \
-      \--test-options='test-tree --output="FILE_NAME"'
+      \--test-options='test-tree --output="output.tmp.json"'
 > cabal test test-suite-name \
-      \--test-options='run-tests --input="FILE" --output="FILE"'
+      \--test-options='run-tests --input="input.tmp.json" --output="output.tmp.json"'
 ```
 
 The `test-tree` command outputs a JSON-serialized `TestTree` to the given
@@ -128,8 +135,71 @@ The `run-tests` command reads a JSON-serialized `RunTestsRequest` from
 
 -------------
 
-For IDEs, they will utilize these commands to find the test tree, execute
-the test tree, and display the results to the user.
+For IDEs, they will utilize these commands to:
+ -  find the test tree
+ -  execute a test run request, and display the results to the user.
+
+One problem that could occur is that having the IDE run `cabal test` to
+reindex the `TestTree` is too slow and incurs too large a latency for
+the user. However, this problem is no worse than the current situation,
+in which the user has to call `cabal test` anyway in order to run the tests.
+
+Furthermore, IDEs often having access to stale artifacts for code that used
+to compile but no longer compiles, therefore, the IDE could run the `test-tree`
+command while still displaying a mostly-correct test tree and inline hints
+in the user's IDE. Once the call to `test-tree` has completed, the new test
+tree can be displayed to the user.
+
+## Alternative Approaches
+
+### A typeclass driven approach with HLS integration
+
+Described in the following [discourse comment](https://discourse.haskell.org/t/hf-coordination-for-test-framework-ide-integration/5221/2?u=santiweight)
+
+In summary, take each function signature and detect whether said function
+is runnable as a test based on whether its type is a member of the
+proposed typeclass `IsTest`. Testing frameworks will each provide their
+own `IsTest` instances for their common testing types.
+
+I have chosen the option in this proposal for the following reasons:
+ 1. Such an approach would run Haskell code in GHCi, which is slower, and
+ would therefore scale poorly for production codebases.
+ 2. Such an approach requires integration with GHC's API, which the author
+ is unfamiliar with and therefore cannot be confident in their ability to
+ maintain the integration library.
+ 3. The author cannot see how to acquire the test tree that is common to
+ IDE test integration, since it requires inspecting the typechecking artifacts
+ of all files in the test suite.
+ 4. Sometimes testing frameworks will provide conflicting `IsTest` instances for the
+ the same type, which is a problem when using multiple testing frameworks in the
+ same codebase.
+
+The typeclass proposal seems incredibly useful however for the typical `Run Program`
+command now provided, or as an additional tool for users. It is certainly a good
+idea for a future project.
+
+### Regex-based test discovery
+
+While regex discovery is a good and quick solution implementation wise, it does
+not scale to many real-world usecases. For example, many codebases will have helpers
+such as:
+
+```haskell
+testDecodeJsonFromFile fileName = testCase fileName $ readJsonFromFile fileName
+```
+
+Since the fileName value is not known until runtime, regex-based approaches miss
+such cases, which is a non-starter for an ecosystem-wide tool.
+
+The proposal put forth by the author will also have to handle this use-case, by
+providing support in testing frameworks for users to define their own combinators
+that declare their location on use site. But this problem is a purely technical
+one with many solutions, such as the following pseudo-code:
+
+```haskell
+testDecodeJsonFromFile :: HasCallStack => TestTree
+testDecodeJsonFromFile = Tasty.customTestCase callstack fileName $ readJsonFromFile fileName
+```
 
 ## Future Work
 


proposals/accepted/50-ide-test-integration.md
```diff
@@ -0,0 +1,219 @@
...
+ ## Background
+
+IDE integration for test suites is a feature that is expected in most
+if not all major programming languages. First class support exists for
+[JUnit with multiple IDEs](https://junit.org/junit5/docs/current/user-guide/#running-tests-ide),
+[Python with VSC](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=&cad=rja&uact=8&ved=2ahUKEwiRyt_VwcD7AhUXIUQIHQe1CjUQFnoECBIQAQ&url=https%3A%2F%2Fcode.visualstudio.com%2Fdocs%2Fpython%2Ftesting&usg=AOvVaw0U1x8oEhgUTX1GeoPJqVqo),
+[.NET with VSC](https://marketplace.visualstudio.com/items?itemName=formulahendry.dotnet-test-explorer)
+and many others.
+
+Often these IDEs will supply inline hints for "Run this test" and
```
david-christiansen, line 23: Here, I think a screenshot would be very useful for Haskellers who don't use IntelliJ much. This one came up from a quick image search, for instance: https://i.stack.imgur.com/j5f1o.png

proposals/accepted/50-ide-test-integration.md
```diff
@@ -0,0 +1,219 @@
...
+"Re-run failed tests" as well as others. For users of mainstream
+languages, such features are no longer features, but instead
+expectations from a mature language.
+
+Haskell has a mature and well-maintained testing ecosystem, including
+the libraries: `tasty`, `HUnit`, and `hspec`.  While these libraries
```
david-christiansen, line 29: What about QuickCheck and friends? Is the intention that they are subsumed by the mention of Tasty here?

proposals/accepted/50-ide-test-integration.md
```diff
@@ -0,0 +1,219 @@
...
+themselves are mature, there is currently minimal
+integration between Haskell's testing libraries and HLS, as well as
+the primary IDEs used for Haskell (VSC;Emacs;Neovim).
```
david-christiansen, line 32:
```suggestion
the primary IDEs used for Haskell (VSC, Emacs, and Neovim).
```

proposals/accepted/50-ide-test-integration.md
```diff
@@ -0,0 +1,219 @@
...
+
+Attempts at integration between testing frameworks have focused on
+regex searches, with support for hspec:
+ 1. [neo-test](https://github.com/mrcjkb/neotest-haskell)
+ 2. [vim-test](https://github.com/vim-test/vim-test)
+
+While these are valuable contributions to the ecosystem, their choice
+of regex for the collection of the user's test tree, and support for only
```
david-christiansen, line 40: The term "test tree" is not necessarily understood by the audience for this proposal . Based on our discussion elsewhere, I think you made a good observation about how Haskell test libraries are typically (but not always) organized, and how it's different from something like JUnit-derived tests.  In particular, JUnit-inspired libraries typically use the language's notion of a declaration or definition to bound tests, so a test is a class or a method, while Haskell libraries more frequently use an ordinary data structure to represent tests. This means that we'd need to be a bit creative about the UI, and draw inspiration from things like IntelliJ while still doing something a bit different that fits Haskell better.

I think the proposal would be accessible to more readers if these assumptions and observations were made more explicit in this background section.

proposals/accepted/50-ide-test-integration.md
```diff
@@ -0,0 +1,219 @@
...
+
+## Goals
+
+1. Create a new library, `hs-test`, with common datatypes for testing framework
+ test trees, and test results.
+2. Generate test trees and test results that can be serialized into JSON from
+Haskell's major testing frameworks, starting with `tasty`.
+3. Integrate test tree output output with VSC's Haskell extension, via VSC's [Testing API](https://code.visualstudio.com/api/extension-guides/testing).
+4. Stretch goal: support IDEs other than VSC.
+5. Stretch goal: support test results caching, and features such as "Rerun only failed tests".
```
david-christiansen, line 71: I can't tell if this goal is a subgoal of number 3, or if it's something that should be supported by the other frameworks. Can you clarify it?

proposals/accepted/50-ide-test-integration.md
```diff
@@ -0,0 +1,219 @@
...
+
+## Technical Proposal
+
+The `hs-test` library will contain the following data declarations:
```
david-christiansen, line 75: I think that being concrete is great here, but you should add a bit of wiggle room in case you learn something in the process. What about:
```suggestion
The `hs-test` library will contain data declarations along the lines of:
```

proposals/accepted/50-ide-test-integration.md
david-christiansen: That is, is it a part of the JSON output format, or is it a VS Code extension feature?

proposals/accepted/50-ide-test-integration.md
```diff
@@ -0,0 +1,219 @@
...
+-- suite that could be run.
+--
+-- `SrcLocs` are used to indicate where inline hints should be displayed
+-- in an IDE.
+data TestTree =
+  TestGroup {name :: Text, subTrees :: [TestTree], location :: SrcLoc}
+  | TestCase {name :: Text, location :: SrcLoc}
+
+data RunTestsRequest =
+  RunTestsRequest {path :: NonEmpty Text}
```
david-christiansen, line 88: Nothing in Tasty guarantees that the names are unique, and if I have my Tasty test tree factored out into a variety of top-level definitions, then I could very easily end up accidentally reusing a name. I would then be surprised if the wrong test is run when I click on it, and it might make me miss a bug.

Is there another way to represent the path?

For instance, what about a redundant encoding of name-position pairs?

proposals/accepted/50-ide-test-integration.md
```diff
@@ -0,0 +1,219 @@
...
+
+-- | A lazy list in order to support arbitrary execution
+-- order and streaming between the test runner and IDE
+data TestResults = TestResults [TestCaseResult]
```
david-christiansen, line 92: Is this a lazy list rather than something in IO to support a hypothetical pure test runner? Does such a thing exist?

proposals/accepted/50-ide-test-integration.md
```diff
@@ -0,0 +1,219 @@
...
+It will also contain the logic to encode this datatype
+to JSON, for consumption by IDEs.
+
+------------
+
+For testing frameworks that want to support IDE integration,
+they should provide support for the following commands:
+
+```bash
+./test-suite test-tree --output="FILE_NAME"
```
david-christiansen, line 117: This is clearly bikeshedding. Sorry!

But I would expect that a protocol designed for machines would not take up an entry in the human-friendly subcommand namespace, and would instead do everything through`--long-options`. And I would really want the output to be able to come over stdout rather than into a file if I were building a front-end for this, because then I don't have to worry about cleaning up temp files and the like.

proposals/accepted/50-ide-test-integration.md
```diff
@@ -0,0 +1,219 @@
...
+data TestCaseResult = TestCaseResult {
+    -- | Represents the path to get to this test case from the
+    -- TestTree, where each element represents a node in the original
+    -- TestTree.
+    path :: NonEmpty Text,
+    result :: TestResult,
+    timeTaken :: Maybe Double
+  }
+
+data TestResult = Success | Skipped | Failure FailureReason
```
david-christiansen, line 103: Having timing information is great. Is it enough?

Using Haskell datatypes as a means of thought is great, but I'd almost prefer the proposal to just define the JSON format. This would allow subtypes to be built (read: with extra fields) to allow the community to experiment with what additional information is useful on top of a common core. For instance, QuickCheck tests might want to show how many tests passed vs were skipped, and that would be useful feedback in my editor.

proposals/accepted/50-ide-test-integration.md
```diff
@@ -0,0 +1,219 @@
...
+Described in the following [discourse comment](https://discourse.haskell.org/t/hf-coordination-for-test-framework-ide-integration/5221/2?u=santiweight)
+
+In summary, take each function signature and detect whether said function
+is runnable as a test based on whether its type is a member of the
+proposed typeclass `IsTest`. Testing frameworks will each provide their
+own `IsTest` instances for their common testing types.
+
+I have chosen the option in this proposal for the following reasons:
+ 1. Such an approach would run Haskell code in GHCi, which is slower, and
+ would therefore scale poorly for production codebases.
```
david-christiansen, line 165: I don't think this is actually true - the type class based approach seems to me like it could support compiled or interpreted mode, and that GHCi is an implementation strategy not forced by the design. Is there something I'm missing?

proposals/accepted/50-ide-test-integration.md
```diff
@@ -0,0 +1,219 @@
...
+ 2. Such an approach requires integration with GHC's API, which the author
+ is unfamiliar with and therefore cannot be confident in their ability to
+ maintain the integration library.
```
david-christiansen, line 168: I think that the design would require the IDE author to know the GHC API - hopefully this would be done once in HLS and then everyone else would get it for free. But the author of e.g. QuickCheck or Tasty would just need to write a four-line instance declaration.

proposals/accepted/50-ide-test-integration.md
```diff
@@ -0,0 +1,219 @@
...
+ 3. The author cannot see how to acquire the test tree that is common to
+ IDE test integration, since it requires inspecting the typechecking artifacts
+ of all files in the test suite.
```
david-christiansen, line 171: This is a true limitation with respect to your proposal. Haskell test libraries really do organize things on a basis other than top-level declarations, and it would be good to not reinvent the wheel. I like that your proposal is highly compatible with what everyone's already using.

proposals/accepted/50-ide-test-integration.md
```diff
@@ -0,0 +1,219 @@
...
+ 4. Sometimes testing frameworks will provide conflicting `IsTest` instances for the
+ the same type, which is a problem when using multiple testing frameworks in the
+ same codebase.
```
david-christiansen, line 174: This is the permanent problem of type classes. Such conflicting instances would always be orphans, right? I think this is no worse than any other Haskell type class.

proposals/accepted/50-ide-test-integration.md
```diff
@@ -0,0 +1,219 @@
...
+The `hs-test` library will contain the following data declarations:
+
+```haskell
+-- | Datatype representing all the tests that exist in a test
+-- suite that could be run.
+--
+-- `SrcLocs` are used to indicate where inline hints should be displayed
+-- in an IDE.
+data TestTree =
+  TestGroup {name :: Text, subTrees :: [TestTree], location :: SrcLoc}
```
david-christiansen, line 84: From where should the testing library get the `SrcLoc`? It's not easy for Haskell programs to view their own source code like this, and I'd be worried to promote an ecosystem-wide convention that requires fragile kludges to be used successfully.

proposals/accepted/50-ide-test-integration.md
```diff
@@ -0,0 +1,219 @@
...
+This proposal is purposefully minimal in scope, for a number
+of reasons:
+ 1. In order for testing frameworks to buy in, we will need a
+ minimally-controversial interface.
+ 2. Support for more features should be done incrementally, since
+ the design space here is rather large, and we will not know users'
+ needs until after integration is alright out in the real world.
+
+Here are some features that have been neglected in this proposal:
+ 1. Distinction between "error" and "failure" when a test fails.
```
david-christiansen, line 214: Why is this not a mere matter of adding an additional constructor to `TestResult`?

proposals/accepted/50-ide-test-integration.md
```diff
@@ -0,0 +1,219 @@
...
+ 2. Support for multiple filters when running a test.
+ 3. Support for options when running a test, such as: `--accept` when
+ running golden tests, `-jN` to run tests in parallel, and options
+ related to property-based tests.
+ 4. Support for localized options when running a test.
```
david-christiansen, line 219: What are localized options? In my head it would be something like replacing `--accept` with `--godtage` to make the Danish localization, but this doesn't seem like a very plausible interpretation of what you're writing.

proposals/accepted/50-ide-test-integration.md
david-christiansen: Also, the JUnit report format is already widely supported. For instance, Gitlab CI will provide detailed test suite summaries if the tests output this format. Why invent a new one here, rather than using the existing standard? (I'm not saying we shouldn't invent a new thing, but the proposal should say why novelty is necessary).

david-christiansen (commented):
I added some general feedback - hope it's useful.

david-christiansen:
One more question: how do we expect the IDE to get from the current filename to the name of the test suite executable to be run? I know how HLS could do this, but the proposal is written to be independent of HLS as far as I can see.

proposals/accepted/50-ide-test-integration.md
```diff
@@ -0,0 +1,219 @@
...
+have gained a valuable IDE tool which will improve the user experience
+of Haskellers as well as make Haskellers more productive.
+
+Support for IDE integration of testing libraries will provide the following
+benefits:
+ 1. Ability for IDEs with Haskell support to provide inline hints for "Run
+ Tests" style commands.
+ 2. Ability for IDEs with Haskell support and support for a test explorer
+ view to display the user's test tree and run individual tests.
+ 3. A unified framework for testing libraries to provide integration with
```
michaelpj, line 55: We should be able to offer a unified CLI interface also, perhaps?

proposals/accepted/50-ide-test-integration.md
```diff
@@ -0,0 +1,219 @@
...
+ IDEs. When a new testing framework is developed, they will be able to
+ provide the results in a Haskell datatype via serialization, which
+ avoids framework-specific test tree discovery logic.
+ 4. Potential to share test result rendering logic between different testing
```
michaelpj, line 59: Up until this point I thought this was mostly about test _discovery_, but now it sounds like you also want to standardise test _reporting_. That might be harder!

proposals/accepted/50-ide-test-integration.md
```diff
@@ -0,0 +1,219 @@
...
+ frameworks, whereas currently each testing framework has their own test result
+ display logic.
+
+## Goals
+
+1. Create a new library, `hs-test`, with common datatypes for testing framework
+ test trees, and test results.
+2. Generate test trees and test results that can be serialized into JSON from
```
michaelpj, line 67: I think you have in mind concepts of "test tree" and "test results", it would really help if you made that explicit. The technical section below does that, but I think it would be helpful to have something in the initial prose, and maybe an argument that it's adequate for all usecases!

proposals/accepted/50-ide-test-integration.md
```diff
@@ -0,0 +1,219 @@
...
+
+```haskell
+-- | Datatype representing all the tests that exist in a test
+-- suite that could be run.
+--
+-- `SrcLocs` are used to indicate where inline hints should be displayed
+-- in an IDE.
+data TestTree =
+  TestGroup {name :: Text, subTrees :: [TestTree], location :: SrcLoc}
+  | TestCase {name :: Text, location :: SrcLoc}
```
michaelpj, line 85: Is there some assumption that given a test in a test tree you can establish how to run that test? It's not clear to me that this gives you that ability.

proposals/accepted/50-ide-test-integration.md
michaelpj: Working out what I imagine would happen here:
- The path gets turned into a tasty pattern
- The pattern matches multiple tests
- Multiple tests get run

Seems somewhat bad but maybe acceptable?

proposals/accepted/50-ide-test-integration.md
michaelpj, line 84: I think we're almost certainly going to want this to be extensible in some way. I think it's very likely that different testing frameworks will want to annotate the tree with framework-specific metadata of their own, and we should make supporting that a key design goal.

michaelpj: I think we definitely want the location to be optional.

proposals/accepted/50-ide-test-integration.md
```diff
@@ -0,0 +1,219 @@
...
+
+`hs-test` will provide helper functions for these serialization tasks.
+
+-------------
+
+For IDEs, they will utilize these commands to:
+ -  find the test tree
+ -  execute a test run request, and display the results to the user.
+
+One problem that could occur is that having the IDE run `cabal test` to
```
michaelpj, line 141: I think having to _actually_ run a cabal command to find this out is a weakness. Junit and friends can know most of this information statically, just by compiling the code, which is quite nice! That does mean that they have to rely on stuff like the hierarchy being implicit in static information (like the class hierarchy), but it is otherwise quite nice. I think we should at least _think_ about whether we can do this _more_ statically (or at least without cabal in the loop!!).

michaelpj: e.g. HLS has some plugins that load code into the interpreter and run it, so we could plausibly compute the test tree without going through cabal, if we do need to run something.

proposals/accepted/50-ide-test-integration.md
michaelpj: I also don't understand this. For running the tests you could still run them through cabal or whatever, assuming that you have some way of knowing how to do that.

michaelpj (commented):
I do think this proposal would be more likely to get adoption if it was more abstract

michaelpj:
I do think this proposal would be more likely to get adoption if it was more abstract. Sketching some ideas:
```
class TestSpecification t where
    type TestResult :: *
    isSuccess :: TestResult t -> Bool
    details :: TestResult t -> Text
    
    type TestKey :: *
    enumerateTests :: t -> [TestKey t]
    runTest :: TestKey t -> IO (TestResult t)

class (TestSpecification t, TestKey t ~ NonEmpty Fragment) => HierarchicalTestSpecification t where
    type Fragment :: *
   
type TestReport t = Map (TestKey t) (TestResult t)

produceTestReport :: (TestSpecification t) => t -> IO (TestReport t)
produceTestReport t = fromList <$> for (enumerateTests t) $ \key -> do
    report <- runTest key
    pure (key, report)
```
This is 5 minutes of thinking, but I think this would give people a lot of flexibility.

proposals/accepted/50-ide-test-integration.md
david-christiansen: This would be a really nice result!

proposals/accepted/50-ide-test-integration.md
david-christiansen: I think that running individual tests from the editor requires both. Discovery allows something like HLS to say "this is a test, want to run it?", and reporting allows HLS to say "It ran, and it passed".

gbaz:
In our last TWG meeting we talked about this a bit. I think there's generally favorable sentiment. One big sticking point that came us is that while mainly "produce a test tree" is a mode existing test frameworks are good at, annotating it with source locations is not. There's an idea that HasCallStack constraints can be used to extend existing frameworks for this in a lightweight way, which would help resolve that. That said, I'd still be onboard with making such locations optional but encouraged.

santiweight:
> In our last TWG meeting we talked about this a bit. I think there's generally favorable sentiment.

Thanks for the response. I am still interested in following up on this, but was ill for the last two weeks.

I think the optional but encouraged nature of the test tree is exactly what we should push for. I do think that producing these locations is not particularly difficult with the HasCallStack.

Thankfully, we can still push for this just with the "test tree" on its own, since that is probably the best way to handle Haskell tests. And it seems clear that we can add optional SrcLoc annotations. So I think I will continue down this road.

> I do think this proposal would be more likely to get adoption if it was more abstract.

I understand this sentiment, and I will consider how to be abstract.

There is a problem with abstractness however, is that a the test-plugin-for-ides level, we suddenly need a single joining point, where all the abstractness becomes concrete so we can display results etc.

I'm happy to go with something abstract, but that, for the time being, every `TestResult t` must be translatable to some `PluginTestResult`. Unfortunately, this removes a lot of the usefulness of a flexible `TestResult`. I'm not sure that maintainers are going to be inclined to use the flexible `TestResult` for uses other than the plugin, and the plugin requires something less abstract.

What use case would something abstract provide if not integration with a plugin?

david-christiansen:
The question about abstraction was intended for @michaelpj , right?

michaelpj:
> One big sticking point that came us is that while mainly "produce a test tree" is a mode existing test frameworks are good at, annotating it with source locations is not.

Perhaps worth pointing out in the proposal that there are two ways of slicing this problem:
1. Run the test harness to discover tests, use this information to locate the tests in the source
2. Interrogate the source to discover tests, use this information to work out how to run them

The characteristic of approach 2 is using some kind of source annotation to identify tests. Approach 2 has the advantage that you don't need to complicate the runtime with source information. AFAICT, approach 2 is _much_ more common in other languages: Java, Rust, Python etc., all have test frameworks that work primarily off source annotations. That's not to say we should do it, but perhaps worth being clear why not. (And we _could_ do it: Haskell has annotations!)

> There is a problem with abstractness however, is that a the test-plugin-for-ides level, we suddenly need a single joining point, where all the abstractness becomes concrete so we can display results etc.

Okay, so what does abstractness get us?

1. Richer test harness behaviour

We are passing this information across a serialization boundary, because we are running the discovery in one process, recording the results, and then feeding that back into the harness. If we serialize to JSON, then in one sense the _real_ interface is that a `TestTree` is a `Value`! But this is much less informative than saying that a `TestTree` is _some type_ specific to the test harness _which can be converted to and from `Value`_. This is nicer for the implementor to begin with, but also since the JSON is being produced and consumed by the same code they can pass through additional information easily:

```
MyTestCase
--> discover
JSON
--> run
MyTestCase
```

So test harnesses can assume they have their own richer `TestTree` type all the time, which is nice.

2. Flexibility

Having an abstract interface that just specifies exactly what we need makes it easier for other people to fit their own system into it. For example, it was only when sketching out an abstract version that I realised that it might be nice to just have a "flat" `enumerateTests` that would work just fine for test frameworks that don't arrange their tests hierarchically. We can then optionally enrich it with additional things if we have a hierarchical arrangement.

 The risk, of course, is that it's over-engineered and nobody ever uses this additional power :shrug: 

> I'm happy to go with something abstract, but that, for the time being, every TestResult t must be translatable to some PluginTestResult.

Right, we need to be able to interpret the test results generically, that's true. This is a rare time when I find myself wishing for subtyping :D 

----

Having written these things, I'm not feeling confident that I'm saying sensible things. I think I'd want to actually try it out before having more opinions :sweat_smile:

santiweight:
Thanks for the response @michaelpj. I neglected to add that discussion about why Haskell is different, which was had in this [discourse comment](https://discourse.haskell.org/t/hf-coordination-for-test-framework-ide-integration/5221/18?u=santiweight). You can also find in various points in that thread discussion around using Haskell's annotation system, which was shot down pretty hard.

> Right, we need to be able to interpret the test results generically, that's true. This is a rare time when I find myself wishing for subtyping :D

I think this is why I want to put off your concerns of abstractions for  follow up to this proposal. A majority of the benefit here is found in simply being able to run Haskell tests in VSC. And we should make sure to design something that is extensible to an abstract design without throwing away implementation work.

All this said: I get the feeling that an implementation is what's needed after this discussion. I will see if I can find the time over this Christmas to get an MVP working :)

gbaz:
@santiweight be sure to check in with @davean who worked up a proof-of-concept of the `HasCallStack` approach that you could leverage.

david-christiansen:
@santiweight How are things going with this proposal? Is there something we can do to get it unstuck?

santiweight:
Hey thanks for checking in. Just finished a three week holiday today. I don't think there's anything blocking me right now.

@gbaz How might I check in with @davean? I don't know what project you're referring to.

My goal for the future is getting an MVP plugin for VSC. I don't think it will be too hard, and I'll surely take a lot of shortcuts.

Some overall thoughts:
 1. I am personally happy with the source location annotations using HasCallStack and will be using that for MVP
 2. I will make a very monomorphic data structure and target exclusively `tasty` test suites
 3. I don't expect an MVP to be too challenging, and mostly about learning VSC's typescript plugins

From the current discussion, I want to implement and iterate. I get the sense that trying to corral Haskell's ecosystem ahead of time is premature (but I expect we will want to do some corralling down the line...).

gbaz:
I think david will send an email connecting you. What he has is basically a proof of concept, and if you already have a HasCallStack proof of concept that works equally well, you may not get much out of his -- at the time I suggested you two connect, it was not clear to me how much (if any) progress you had made in that regard. I look forward to seeing what you come up with!

santiweight:
https://user-images.githubusercontent.com/16826028/213981133-70ef0cc5-37d8-443d-92c1-d19f75f8c676.mov

santiweight:
There is a small wrinkle that I'm not sure we can overcome without enabling deep subsumption.

In order to have a location attached to the correct source location, we need to have a `HasCallStack` in scope for the declaration containing a test definition.

In order to achieve this, we could define:

```haskell
data TestTree = TestCase | TestGroup
type TestTree = HasCallStack => TestTree
```

It's not ideal imo. I'm open to suggestions here.

santiweight:
I also have locations in the text working...

https://user-images.githubusercontent.com/16826028/213982513-9b2043a2-f3ee-4cf2-bd45-8e0acb9c98c2.mov

michaelpj:
Nothing specific to say except that looks super cool!

santiweight:
Help needed and welcome!

There are a few tasks left to do that I don't know how to resolve.

1. I need to find a way to get the available test projects in a VSC workspace. Subsequently I need to add better caching based on the files in a test project. Is this what `hie-bios` is for?
2. Preferably we can find a way to avoid the need for a `HasCallStack` constraint on the function that calls `testCase` or `testGroup`. Above I suggested a solution that requires `DeepSubsumption` which is not satisfactory imo...

Any input on either of these? If not then I'll just go and put out a crappy first try version.

fendor:
Without having read the full context (sorry, just saw your question re hie-bios), hie-bios's main purpose is to find the compilation option given a filepath or some other way to identify a part of your project. Any caching it does is for performance reasoning.
HLS caches these components, and should know all the files it has loaded. What caching are you referring to in particular?

santiweight:
Thanks @fendor. I misspoke in my comment. I meant "file-watching" - so that I know when a test tree becomes stale. I'm not sure the best way to find either what test projects are available in a folder (they might not be at the top level of the folder), nor do I know how to do file-watching for a cabal project.

I'm assuming both of these use cases have been paved?

fendor:
> I'm not sure the best way to find either what test projects are available in a folder

For a cabal project, I think there is no built-in way to do it, unless something like https://github.com/haskell/cabal/pull/7500 lands. 
You can find the components via a query such as `cat dist-newstyle/cache/plan.json | jq '."install-plan"|.[]|select(.style=="local")|select(."component-name" != null) | select(."component-name" | startswith("test:"))|."component-name"'`, which basically just queries `plan.json` for any local component whose `component-name` starts with `test:`. (or using cabal-plan directly)

> nor do I know how to do file-watching for a cabal project

Personally, also not sure. I have some ideas, but I don't know of a way to do it right now.

santiweight:
I tried to come back to this because I think it's really cool, but I don't know the way forward with respect to file watching.

Does anyone have any clue who we could reach out to about this?

It would be a shame to have this die, because it does work, and we're not terribly far away from an MVP.

david-christiansen:
Well, I seem to have missed this final question - sorry. For file watching, I'd post the question on Discourse - I don't know a more specific forum to check.

I'm going to temporarily close this one - please feel free to reopen it if you want to keep the discussion going and make progress!

PR #50, opened 2023-06-16
Proposal to specify JSON dump in GHC

benbellick:
GHC's `-ddump-json` is underspecified. This is a proposal to officially support a JSON dump of GHC's diagnostic messages. The purposed of this proposal is to garner feedback from stakeholders, as well as broadcast the effort to a larger community in order to better understand the level of desire as well as feature requirements. 

[Rendered](https://github.com/benbellick/tech-proposals/blob/specify-json-dump/proposals/0000-specify-dump-json.md)

[Proposed Schema](https://github.com/benbellick/tech-proposals/blob/specify-json-dump/proposals/0000-specify-dump-json/schema.json)

### DIFF
--- a/proposals/0000-specify-dump-json.md
+++ b/proposals/0000-specify-dump-json.md
@@ -2,7 +2,7 @@
 
 ## Abstract
 
-GHC is currently undergoing development GHC is currently undergoing a long scale project to [move to a more structured error representation](https://gitlab.haskell.org/ghc/ghc/-/issues/18516). 
+GHC is currently undergoing a long term project to [move to a more structured error representation](https://gitlab.haskell.org/ghc/ghc/-/issues/18516). 
 In harmony with this motion, there is a [desire for a JSON dump of GHC's diagnostic messages](https://gitlab.haskell.org/ghc/ghc/-/issues/19278).
 Such an implementation would facilitate easier consumption for downstream consumers of GHC's diagnostic messages (e.g. IDE developers). 
 To best enforce a standardized JSON output, a [JSON schema](https://json-schema.org/) can be used.
@@ -66,7 +66,7 @@ The extra `d` in `-ddump-json` indicates that this is a developer flag, but our 
 
 The aim is to create a JSON interface for GHC diagnostics, standardized by a JSON schema. Then, the `-ddump-flag` flag will be replaced by the first-class `-dump-flag` flag, and subsequently, whatever implementation changes are necessary to make the new flag conform to the new specification will be performed.
 
-The past attempt to not succeed simply due to lack of effort and resources devoted to it, as well as a lack of structured error representations. Now that structured errors are pervasive within GHC, such a JSON dump is both feasible and practical.
+The past attempt did not succeed simply due to lack of effort and resources devoted to it, as well as a lack of structured error representations. Now that structured errors are pervasive within GHC, such a JSON dump is both feasible and practical.
 
 ## Technical Content
 
@@ -134,7 +134,7 @@ One of the major benefits of utilizing a JSON schema is that the expected JSON p
 
 In addition to adding a `-dump-json` flag, it may also prove useful to provide a `-dump-json-schema` flag which simply produces the relevant JSON schema for that particular version of GHC. This I leave open for discussion. Provided that the schema is in an easy to find location, it may be overkill.
 
-The schmema evolution process is currently undetermined, though I imagine that due to the infrequency with which the schema will need to be changed, it can be handled on a case-by-base basis. Though keeping a running list of all relevant stakeholders that may need to be informed could be a good idea. 
+The schema evolution process is currently undetermined, though I imagine that due to the infrequency with which the schema will need to be changed, it can be handled on a case-by-base basis. Though keeping a running list of all relevant stakeholders that may need to be informed could be a good idea. 
 
 ## Stakeholders
 
--- a/proposals/0000-specify-dump-json.md
+++ b/proposals/0000-specify-dump-json.md
@@ -119,10 +119,7 @@ For demonstrative purposes, here is an example valid instance of the schema.
       },
       "severity": "Error",
       "code": 27958,
-      "message": "    • Couldn't match type ‘b0’ with ‘(Bool, b0)’ \n Expected: b0 -> Maybe (Bool, b0) \nActual: b0 -> Maybe b0 \n• In the first argument of ‘unfoldr’, namely ‘Just’ \nIn the expression: unfoldr Just \nIn an equation for ‘f’: f = unfoldr Just",
-      "warnReason": {
-        "reason": "ErrorWithoutFlag"
-      }
+      "message": "    • Couldn't match type ‘b0’ with ‘(Bool, b0)’ \n Expected: b0 -> Maybe (Bool, b0) \nActual: b0 -> Maybe b0 \n• In the first argument of ‘unfoldr’, namely ‘Just’ \nIn the expression: unfoldr Just \nIn an equation for ‘f’: f = unfoldr Just"
     }
   ]
 }
--- a/proposals/0000-specify-dump-json.md
+++ b/proposals/0000-specify-dump-json.md
@@ -21,8 +21,8 @@ These `SDoc`s are simply a text encoding of the errors with some formatting feat
 By transforming the errors into `SDoc`s at the moment of creation, a lot of type-level information is lost,
 causing a headache for GHC developers and GHC API users.
 To improve on this problem, the [journey to represent errors as structured values has begun](https://gitlab.haskell.org/ghc/ghc/-/wikis/Errors-as-(structured)-values ). 
-In summary, this means to pass errors around as a specific type which contains the relevant information separated into different fields,
-rather than combining it all into one single `SDoc`. 
+In summary, this means passing errors around as a specific type which contains the relevant information separated into different fields,
+rather than combining it all into one single `SDoc`.
 
 Another change that has been made is the introduction of [error codes](https://gitlab.haskell.org/ghc/ghc/-/issues/21684) to uniquely identify each possible error output from GHC. 
 
@@ -34,9 +34,9 @@ The fundamental issue is that the structured errors are only presented to users 
 
 The first is that this is an _unnecessary_ pain. The errors are now passed as structured values internally, but the users are given a fairly unstructured value to parse themselves, leading to the requirement of always having to write a parser whenever consumption is desired.
 
-The second problem is that this plain text output is subject to change. Messages are often changed to make the content clearer, but each change results bubbles into a new requirement for the hand-written plain-text parser of error messages. This problem is mitigated a bit by the above-mentioned new inclusion of error codes, but even these error codes must be parsed from a non-standard structured output. 
+The second problem is that this plain text output is subject to change. Messages are often changed to make the content clearer, but each change bubbles into a new requirement for the hand-written plain-text parser of error messages. This problem is mitigated a bit by the above-mentioned new inclusion of error codes, but even these error codes must be parsed from a non-standard structured output. 
 
-The requirements to solve this problem are to enable the consumption of GHC's diagnostic message in a structured representation that closely matches the internal representation without requiring users to utilize the internals of GHC. The most reasonable solution is to use a compiler flag to dump the structured messages into a versioned JSON output, which is described by a JSON Schema, to standardize the expected output. The hope is that the community can reach a general consensus on the requirements imposed by such a JSON schema. While it will not be too costly to change the JSON schema as new requirements come up (simply increment the schema version), the ideal situation is to present here an opportunity to provide input on the current schema and receive constructive criticism. The conversation is also open on whether or not a JSON schema is really required.
+The requirements to solve this problem are to enable the consumption of GHC's diagnostic message in a structured representation that closely matches the internal representation without requiring users to utilize the internals of GHC. The most reasonable solution is to use a compiler flag to dump the structured messages into a versioned JSON output, which is described by a JSON Schema, to standardize the expected output. The hope is that the community can reach a general consensus on the requirements imposed by such a JSON schema. While it will not be too costly to change the JSON schema as new requirements come up (simply increment the schema version), the ideal situation is to present here an opportunity to provide input on the current schema and receive constructive criticism.
 
 ## Prior Art and Related Efforts
 
@@ -64,13 +64,17 @@ Just on the surface, the first object is of a different kind than the second, th
 
 The extra `d` in `-ddump-json` indicates that this is a developer flag, but our aim is to make this a first class feature.
 
-The aim is to create a JSON interface for GHC diagnostics, standardized by a JSON schema. Then, the `-ddump-flag` flag will be replaced by the first-class `-dump-flag` flag, and subsequently, whatever implementation changes are necessary to make the new flag conform to the new specification will be performed.
+The aim is to create a JSON interface for GHC diagnostics, standardized by a JSON schema. Then, the `-ddump-flag` flag will be replaced by the first-class `-dump-diags-json` flag, and subsequently, the implementation will be changed so that the diagnostics will be output in a JSON format which conforms to the schema presented in this proposal.
 
 The past attempt did not succeed simply due to lack of effort and resources devoted to it, as well as a lack of structured error representations. Now that structured errors are pervasive within GHC, such a JSON dump is both feasible and practical.
 
 ## Technical Content
 
-The most important component of the technical content is the JSON schema itself. The JSON schema is not reproduced here, as it is quite long. Instead, it can be found in the directory of the same name, with title `schema.json`. To demonstrate the appropriateness of the proposed JSON schema, the internal error representation is produced below. Comments removed for brevity.
+The most important component of the technical content is the JSON schema itself. The JSON schema is not reproduced here, as it is quite long. Instead, it can be found in the directory of the same name, with title `schema.json`. 
+
+The schema consists of some top-level information, as well as a list of diagnostic messages. The top level information is simply a `version` field and a `ghcVersion` field. The `version` field is a crucial way for users to evaluate changes to the JSON schema. It will be updated each time a change is made to the JSON schema. The `ghcVersion` field reports the version of GHC. This is useful information because components of the JSON output may want to expose JSON-ified internals of GHC which may themselves have output dependent on GHC's version. Additionally, the GHC version serves as a marker to indicate the progress of error code coverage for the ongoing effort to supply error codes for all diagnostic messages. Supplying this information allows consumers to be sensitive to those changes.
+
+What follows is an explanation of the diagnostic message's schema. To demonstrate the appropriateness of the proposed schema, the internal diagnostic representation is produced below. Comments are removed for brevity.
 ```haskell
 -- compiler/GHC/Types/Error.hs
 data MsgEnvelope e = MsgEnvelope
@@ -84,7 +88,6 @@ class (HasDefaultDiagnosticOpts (DiagnosticOpts a)) => Diagnostic a where
   type DiagnosticOpts a
   diagnosticMessage :: DiagnosticOpts a -> a -> DecoratedSDoc
   diagnosticReason  :: a -> DiagnosticReason
-
   -- | Extract any hints a user might use to repair their
   -- code to avoid this diagnostic.
   diagnosticHints   :: a -> [GhcHint]
@@ -94,20 +97,25 @@ All of the above is explained in the [wiki](https://gitlab.haskell.org/ghc/ghc/-
 
 The output of the JSON dump would consist of a list of errors, as well as a top level version. Each error contains the following fields:
  - `"span"` corresponding to the above `SrcSpan`
+   - required by the schema
    - This contains sub-fields, specified in the schema, which contains line numbers, etc.
  - `"severity"` corresponding to the above `Severity`
+   - required by the schema
  - `"code"` corresponding to the change in this [proposal](https://github.com/haskellfoundation/tech-proposals/blob/main/proposals/accepted/024-error-messages.md) and the above `DiagnosticCode` in typeclass `Diagnostic`
+   -  this is required by the schema but is allowed to have a value of `null`. The hope is that once all diagnostics officially have error codes, this can be required and nonnullable, but that will happen once `diagnosticCode` itself doesn't return a `DiagnosticCode` wrapped in a `Maybe`
  - `"hints"` corresponding to the above `[GhcHint]`
-   - This is perhaps one that remains a bit of a question mark. To what degree should this be specified? There are quite a few possible constructors, and so it may be overly tedious to specify every single one as a possible output in the JSON schema. On the other hand, it may be insufficient to allow arbitrary JSON to be put forward. Maybe a reasonable compromise is to give back the name of the constructor followed by a list of the constructors arguments. 
+   - The schema itself doesn't validate at all what these may be, only that they are themselves objects. In my view, the best way to handle them is to output a JSON-ified version of the type `GhcHint`. Any attempt to specify them more than that will result in spiraling complexity and a required update every time the `GhcHint` type changes. Another alternative is to just remove them from the output and only present suggestions as they are rendered in the `message`.
  - `"message"` corresponding to `DecoratedSDoc` of `diagnosticMessage`
- - `"reason"` corresponding to `diagnosticReason`
+   - this is required by the schema, and is the actual string output produced at the command line
  
- The top level version is crucial for indicating to downstream consumers which JSON schema they must comply with in the case that the schema is updated in the future (which is quite likely).
- 
+The top level version is crucial for indicating to downstream consumers which JSON schema they must comply with in the case that the schema is updated in the future (which is quite likely).
+
+Information encoded in `diagnosticReason` is not included in this initial version, as that feature is currently unstable. It can be incorporated down the line once it is both more stable and there is a desire to consume that information from users. 
 For demonstrative purposes, here is an example valid instance of the schema.
 ```json
 {
   "version": "1.0.0",
+  "ghcVersion": "9.2.8",
   "diagnostics": [
     {
       "span": {
@@ -138,8 +146,9 @@ The schema evolution process is currently undetermined, though I imagine that du
 Stakeholders consist of anyone that desires to consume a JSON output of GHC diagnostics without leveraging the GHC API. Some possible stakeholders are listed explicitly [here](https://gitlab.haskell.org/ghc/ghc/-/issues/19278#note_503994). These are:
  - Chris Smith of https://code.world
  - Joseph Sumabat
+ - Tom Smeding of https://play.haskell.org 
 
-Joseph Sumabat replied to my emails and his input has been incorporated into the included schema. I believe the Haskell community would benefit as a whole from improved tooling, and this effort will help in this regard by making an easy to consume representation of GHC's diagnostics.
+I believe the Haskell community would benefit as a whole from improved tooling, and this effort will help in this regard by making an easy to consume representation of GHC's diagnostics.
 
 ## Success
 
--- a/proposals/0000-specify-dump-json.md
+++ b/proposals/0000-specify-dump-json.md
@@ -104,6 +104,7 @@ The output of the JSON dump would consist of a list of errors, as well as a top 
  - `"code"` corresponding to the change in this [proposal](https://github.com/haskellfoundation/tech-proposals/blob/main/proposals/accepted/024-error-messages.md) and the above `DiagnosticCode` in typeclass `Diagnostic`
    -  this is required by the schema but is allowed to have a value of `null`. The hope is that once all diagnostics officially have error codes, this can be required and nonnullable, but that will happen once `diagnosticCode` itself doesn't return a `DiagnosticCode` wrapped in a `Maybe`
  - `"hints"` corresponding to the above `[GhcHint]`
+   - While these are required by the schema, `[]` is an allowed value
    - The schema itself doesn't validate at all what these may be, only that they are themselves objects. In my view, the best way to handle them is to output a JSON-ified version of the type `GhcHint`. Any attempt to specify them more than that will result in spiraling complexity and a required update every time the `GhcHint` type changes. Another alternative is to just remove them from the output and only present suggestions as they are rendered in the `message`.
  - `"message"` corresponding to `DecoratedSDoc` of `diagnosticMessage`
    - this is required by the schema, and is the actual string output produced at the command line
@@ -111,6 +112,7 @@ The output of the JSON dump would consist of a list of errors, as well as a top 
 The top level version is crucial for indicating to downstream consumers which JSON schema they must comply with in the case that the schema is updated in the future (which is quite likely).
 
 Information encoded in `diagnosticReason` is not included in this initial version, as that feature is currently unstable. It can be incorporated down the line once it is both more stable and there is a desire to consume that information from users. 
+
 For demonstrative purposes, here is an example valid instance of the schema.
 ```json
 {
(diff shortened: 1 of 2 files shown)


proposals/0000-specify-dump-json.md
```diff
@@ -0,0 +1,149 @@
...
+
+In addition to adding a `-dump-json` flag, it may also prove useful to provide a `-dump-json-schema` flag which simply produces the relevant JSON schema for that particular version of GHC. This I leave open for discussion. Provided that the schema is in an easy to find location, it may be overkill.
+
+The schmema evolution process is currently undetermined, though I imagine that due to the infrequency with which the schema will need to be changed, it can be handled on a case-by-base basis. Though keeping a running list of all relevant stakeholders that may need to be informed could be a good idea.
```
adamgundry, line 137:
```suggestion
The schema evolution process is currently undetermined, though I imagine that due to the infrequency with which the schema will need to be changed, it can be handled on a case-by-base basis. Though keeping a running list of all relevant stakeholders that may need to be informed could be a good idea. 
```

proposals/0000-specify-dump-json/schema.json
```diff
@@ -0,0 +1,115 @@
...
+              "reason": {
+                "description": "The reason why a diagnostic was emitted in the first place (e.g., flag, category)",
+                "type": "string",
+                "enum": [
+                  "WarningWithoutFlag",
+                  "WarningWithFlag",
+                  "WarningWithCategory",
+                  "ErrorWithoutFlag"
+                ]
+              },
```
adamgundry, line 55: I'm not convinced the `DiagnosticReason` belongs in the schema as-is. This type has changed quite a bit recently (`WarningWithCategory` is very new, and indeed current HEAD has `WarningWithFlags` rather than `WarningWithFlag`). Is it really useful to consumers? If so, we should think about a more sensible format I think.

proposals/0000-specify-dump-json.md
```diff
@@ -0,0 +1,149 @@
...
+}
+```
+
+Just on the surface, the first object is of a different kind than the second, they are not wrapped in array, and the second object lacks crucial information like error codes. And what is messageClass? These fields can and should be standardized to most authentically express the diagnostics in a way that makes them understandable and consumable.
+
+The extra `d` in `-ddump-json` indicates that this is a developer flag, but our aim is to make this a first class feature.
+
+The aim is to create a JSON interface for GHC diagnostics, standardized by a JSON schema. Then, the `-ddump-flag` flag will be replaced by the first-class `-dump-flag` flag, and subsequently, whatever implementation changes are necessary to make the new flag conform to the new specification will be performed.
+
+The past attempt to not succeed simply due to lack of effort and resources devoted to it, as well as a lack of structured error representations. Now that structured errors are pervasive within GHC, such a JSON dump is both feasible and practical.
```
adamgundry, line 69:
```suggestion
The past attempt did not succeed simply due to lack of effort and resources devoted to it, as well as a lack of structured error representations. Now that structured errors are pervasive within GHC, such a JSON dump is both feasible and practical.
```

adamgundry (commented):
Thanks for this proposal! It seems like fundamentally a good idea to have a convenient machine-readable format for GHC's error messages, and JSON Schema is a sensible way to specify it.

One thing I'm surprised about is that the diagnostic itself is represented as a string, rather than anything more structured. While including the string is sensible, I can imagine tools wanting to analyse the content. Of course there is a challenge here in that the types representing possible diagnostics have many constructors and are subject to change, so it isn't obvious how much they should be baked into the schema. But we could at least have a field containing a JSON-serialised version of the internal GHC diagnostic, and perhaps we could even generically derive the schema itself (so each GHC version could ship with its particular schema).

proposals/0000-specify-dump-json/schema.json
benbellick: This is something I was thinking as well. I included it by default as it is part of the specification of the diagnostic messages, but I am open to leaving it out, especially if it is subject to change in the immediate future. 

Snooping around at the blame, it appears you are one of the implementors of this feature, which makes you particularly suited to judge this feature's stability 😄 

I see no fundamental issue with leaving it out, and then adding it back in later if it is a requested feature by someone.

proposals/0000-specify-dump-json/schema.json
```diff
@@ -0,0 +1,115 @@
...
+            "type": [
+              "integer",
+              "null"
+            ]
+          },
+          "hints": {
+            "description": "The hints suggested by GHC",
+            "type": "array"
+          },
+          "message": {
```
benbellick, line 38: > One thing I'm surprised about is that the diagnostic itself is represented as a string, rather than anything more structured. While including the string is sensible, I can imagine tools wanting to analyse the content. Of course there is a challenge here in that the types representing possible diagnostics have many constructors and are subject to change, so it isn't obvious how much they should be baked into the schema. But we could at least have a field containing a JSON-serialised version of the internal GHC diagnostic, and perhaps we could even generically derive the schema itself (so each GHC version could ship with its particular schema).

To make sure I am understanding you correctly, you mean that this particular field could be a JSON serialized version of the diagnostic message, rather than the message text itself? That is a great point. I believe that that should be added in addition to the text version. 

Any stringent requirements dictated on the schema on the shape of the data in such a field does seem really difficult to maintain, especially considering that the "errors-as-values" proposal is yet to be fully implemented. Agreed, that having a field that contains the JSON-serialized error with no schema constraint is a good idea. Thanks for the suggestions!

LaurentRDC:
I think that's a great proposal!

[I agree with the sentiment that the `message` property shouldn't just be a string](https://github.com/haskellfoundation/tech-proposals/pull/50#pullrequestreview-1484266992). However, since the purpose of this proposal is to allow 3rd party tools to consume error reports *without* depending on GHC internals, I don't think that making the `message` property a JSON-serialization of some GHC-specific data structure the right approach either. 

Instead, why not re-use a light markup language instead? Perhaps a combination of breaking every `message` into an array of lines, each formatted using Markdown. If there is a strong desire to have the serialized `message` structure as well, maybe the JSON `message` field could contain two fields: a list of marked-up lines, and a JSON-encoded GHC diagnostic. I'm not familiar with JSON Schema, but I'm thinking the example in your proposal would look like this:

```json
{
  "version": "1.0.0",
  "diagnostics": [
    {
      (... snip ...)
      "message": {
        "markup": [
          "* Couldn't match type `b0` with `(Bool, b0)` ",
          "Expected: `b0 -> Maybe (Bool, b0)`",
          "Actual: `b0 -> Maybe b0`",
          "In the first argument of `unfoldr`, namely `Just`",
          "In the expression: `unfoldr Just`",
          "In an equation for `f`: `f = unfoldr Just`"
        ],
        "data": "< JSON representation >"
      },
      "warnReason": {
        "reason": "ErrorWithoutFlag"
      }
    }
  ]
}
```

benbellick:
@LaurentRDC I see the appeal of including some _markup-like_ language, though I worry that it may introduce more complexity than benefit. What use-case to you foresee for the _markup_-ified diagnostic string that cannot be fulfilled by just outputting the string? 

As for the inclusion of the JSON internals, I can see an argument for both sides. Some of GHC's errors are quite complex, and they would be made even more complex to end users by exposing the JSON representation of the entire error. 

Perhaps a simpler approach is to toss out including a JSON version of the error value and just use the string. This is closer in spirit to how [rust](https://doc.rust-lang.org/rustc/json.html) seems to do it. The hope is that downstream users rely on the error code and the span for consuming diagnostics, rather than hand parsing a string / markup output, or using the possibly unreliable internals of GHC's diagnostic messages.

LaurentRDC:
> What use-case to you foresee for the _markup_-ified diagnostic string that cannot be fulfilled by just outputting the string?

In my mind, the error message could contain three kinds of markup: raw text, code, and hyperlinks. I don't know if any GHC error message currently contains hyperlinks though.
In any case, users or tools can easily figure out which error.haskell.org page is associated with an error using the error code. You might be right that markup is overkill.

david-christiansen:
While Markdown is a perfectly reasonable input format for text boxes like the one I'm currently typing in, and is much nicer to use than BBCode or similar, it strikes me as an odd _output_ format for machine-to-machine communication. In particular, JSON is nice because most languages have at least one decent parser for it, and it's not hard to write an only-somewhat-broken one in an afternoon. But Markdown exists in a variety of dialects, some better-specified than others, and the quality of Markdown parsers varies much more. It's also surprisingly difficult to actually implement Markdown. Using Markdown as a machine interchange format means that people now have to write (or locate on their favorite package server) two parsers instead of one.

Additionally, GHC's output can contain much more information than can be comfortably represented in Markdown. For instance, source spans are rendered to strings as carats and line positions, and types are separate from programs. Markdown just allows you to say "code" or "not code".

Clearly, the string should be there - asking clients of this information to write their own error pretty-printer is unreasonable. What about a compromise that passes metadata about sub-regions of the string along with the string? E.g.
```
{ "messageText": "<interactive>:1:1: error: Variable not in scope: x",
  "messageInfo": [{"type": "sourcePosition", "start": 0, "length": 18}, {"type": "errorCategory", "start": 20, "length": 5, "info": {"severity": "error"}}, ...]
}
```
With a further invariant that the described regions may contain others, but that they form a tree. That makes it fairly easy for clients to give whatever rendering they need, for them to start simple and slowly get fancier over time, etc. It also allows a fairly simple JSON schema that's not going to change radically at each compiler release.

tek:
> I don't know if any GHC error message currently contains hyperlinks though.

The "noncanonical method" errors contain links.

david-christiansen:
Now that I've got my nitpick hat off, this seems to me to be a project that can bring real value.

I like that you've manually designed a format for the top-level information. That seems to me to be more likely to succeed than using something automatically generated from GHC's internals, and there's always space to add more information in a future revision as things settle down. One way that this could work would be to have two version fields: one for the overall format which is fairly stable, and one for the GHC version in particular, and generically derive a description of the guts for that particular version. Then, clients who want to work at a high level and have a stable tool can look only at the container version, while clients who want to work more intensely with specific error message details can look at the specific GHC point release version.

I worry a bit about the flag namespace. Is `dump-json` the right name for something that dumps only errors and warnings to JSON? I can imagine that a standard JSON serialization of Core, or of type checker traces, or of renamer output could also be useful for various tasks. Do you conceive of this as a flag that makes all compiler output into JSON in the long run, or should there be separate JSON dump flags for other aspects?

It seems to me that this would also be very useful for play.haskell.org - have you been in touch with Tom Smeding to check that the format could be useful there?

Just to be 100% clear: this proposal is a community RFC, not a request for HF funding, right?

Finally, if you're up for it, I think that a minimal library on Hackage that parses this output could really increase the impact of the feature.

Great stuff!

benbellick:
I just sent an email to Tom Smeding. Thanks for mentioning him!

First, the inclusion of a field for the version for the schema as well as a field for the version of GHC seems useful. I imagine it especially useful to those who are developing tools that would hope to cover a variety of GHC versions, rather than being locked into one. 

Second, agreed that the namespace may pose an issue. I can imagine a circumstance in which arbitrary GHC outputs may want to be consumed in a JSON output (e.g. AST, typechecker output), but that will introduce much much more complexity. Instead, lets call the command something like `-dump-diags-as-json`. It is slightly verbose, but also clear in what it is doing. If down the line there is desire for a more general flag which allows anything else to be dumped as JSON, we can make the change then. Also, this output is meant to be consumed by users of GHC, rather than developers of GHC (hence the dropping of the double d). To me, this makes an even stronger case for giving a separate flag for JSON dumps of error diagnostics.

Regarding funding, I am actually doing this project via [funding from Summer of Haskell](https://summer.haskell.org/news/2023-05-14-summer-of-haskell-2023-project-selections.html#standardize-ghcs-error-dump-in-json-format). So yes, this proposal is technically just a request for feedback, though the Haskell Foundation would be funding this work through a different program. 

I am open to (and would be excited to!) write a library which parses the output. However, that is an addition I would rather make provided I have the time. I would like to get the core of the proposal in first. After that, I am happy to open that discussion up again!


Regarding this idea: 
> Clearly, the string should be there - asking clients of this information to write their own error pretty-printer is unreasonable. What about a compromise that passes metadata about sub-regions of the string along with the string? E.g.
> 
> ```
> { "messageText": "<interactive>:1:1: error: Variable not in scope: x",
>   "messageInfo": [{"type": "sourcePosition", "start": 0, "length": 18}, {"type": "errorCategory", "start": 20, "length": 5, "info": {"severity": "error"}}, ...]
> }
> ```
> 
> With a further invariant that the described regions may contain others, but that they form a tree. That makes it fairly easy for clients to give whatever rendering they need, for them to start simple and slowly get fancier over time, etc. It also allows a fairly simple JSON schema that's not going to change radically at each compiler release.

While I also think this could be a good idea, I worry about over-engineering the solution. Maybe a better practice is to just give the string in a first iteration and then see if there is a desire from consumers to have more structure later. I'd like to keep the maintenance cost as low as possible while also providing maximal benefit. So then the real question is, would anyone benefit from this feature now? I have not *yet* heard someone ask for more than 
 - the range
 - error message
 - diag severity
 - message string output
 - error code

but if someone does request more (i.e. structured JSON of message string output), then I can see a stronger case for adding that feature. 

Thanks for all of the feedback and conversation so far!

tomsmeding:
My two cents here as a generic Haskell programmer and maintainer of play.haskell.org (though I have precious little time for the latter at the moment, unfortunately) -- thanks Christian :)

First some random votes:
- I'd take something more specific than `-dump-json` too; personally I'd take `-dump-diags-json` because it's dumping the diags, and the format is JSON. But a careful perusal of the existing GHC flags might uncover that my feeling for a sensible naming pattern is unconventional in GHC's flags.
- Vote against markdown for the same reasons as Christian.

I feel that most consumers of these structured diagnostics would be interested in a _stable_ interface (i.e. backwards-compatible additions would be fine, incompatible changes would be annoying). Beyond that requirement, needs differ: the playground, or an IDE, might want to somehow link diagnostics to the place in the program where they occurred and possibly offer to apply hints contained therein. Analysis tools (I'm imagining some tool that runs in CI and collects some statistics on produced diagnostics) may be very interested in precisely what diagnostic is being produced for what reason, whereas IDEs might not bother much, assuming that GHC formats its diagnostics decently. Perhaps you can think of other clients.

As always, including a version field is essential, but even having one it would be nice not to have to change that version field very much. (Or if it's semver -- I'm not familiar with JSON schema, sorry -- to not have to change the major version component very much.) For IDE/playground usecases, I would imagine that developers would incrementally add functionality to their diagnostic parsing, starting with only the basics (span, severity, message), later expanding (i.e. offering their users fancier functionality) to other components of the diagnostic as developer time allows, users demand, or the schema specifies.

As a result, I would refrain from specifying anything in the schema that is expected to be in flux.
- Surely compiler diagnostics will contain spans and severities forever, and additions to the list of severities will be naturally rare. Also include the error code: it's the most structured part of the structured diagnostics!
  - Consider making the `file` component of a span an absolute path. This is overkill for human-readable error messages, but a boon for robustness for tooling. It seems to me like this structured format is _the_ opportunity to introduce that. (The playground won't care, but others will.)
- Also please **include the formatted message as-is as a string**, to make initial implementation of smart diagnostics parsing doable instead of a large task of parsing the whole schema. Some time in the future, when more parts of the diagnostics can be stably specified, add fields for that but leave the message string intact.
- Following my above advice, I would **refrain from under-specifying hints**. If they're not stable, or you aren't sure on the best way to represent them, don't. Or allow some way to add unspecified fields to the output, where you can then put experimental stuff -- that's fine.
  - Before adding hints to the schema, I would reflect on why people might want hints in the JSON diagnostics. If only some are specified, or if they are included as plain strings, or as GHC constructor names (which are surely unstable!), then they aren't terribly useful for IDE quickfixes -- and I cannot think of another client who would like to use them.

Further small remarks:
- Again, I don't know JSON schema, but it looks like you're allowing for the diagnostic code to be _both_ null and absent -- the field is not in the `"required"` list. Please make it either nullable _or_ non-required.
- I subjectively dislike optional fields in JSON. Make them nullable instead, or use the natural zero value of the type if that makes semantic sense, such as an empty array for hints. This will result in lazy developers (who will never ever look at such a thing as a schema) having a somewhat higher probability of writing code that doesn't break down completely when a diagnostic arises that contains a field they hadn't seen yet.

I like the idea :)

benbellick:
I altered the specification to require hints to be objects. Maybe the best practice is to simply JSON-ify the `GhcHint` type. It seems too complex to have to separately manage a schema for hints, especially considering that specifying all hints within the schema would couples updates to the schema with updates to `GhcHint` in GHC. Instead, I propose either giving the JSON-ified version of `GhcHint`, or just not supplying it at all. 

FYI, I have made the `code` field of the diagnostic message required, but it can be null. The reason for that is to account for those error messages which do not yet have an associated error code. Eventually, that field will be required and non-nullable.

Lastly, I don't quite understand the benefit of making the `file` field show an absolute path. What is the harm in making it a relative bath with root directory the directory in which GHC is called? That is how the current `-ddump-json` does it.

tomsmeding:
@benbellick In retrospect you're right on all points in your last post I think. Use a relative path, not absolute, lest tooling gets confused in the presence of symlinks and varying normalisation routines and whatnot.

proposals/0000-specify-dump-json.md
```diff
@@ -0,0 +1,155 @@
...
+These `SDoc`s are simply a text encoding of the errors with some formatting features, such as nice alignment.
+By transforming the errors into `SDoc`s at the moment of creation, a lot of type-level information is lost,
+causing a headache for GHC developers and GHC API users.
+To improve on this problem, the [journey to represent errors as structured values has begun](https://gitlab.haskell.org/ghc/ghc/-/wikis/Errors-as-(structured)-values ). 
+In summary, this means passing errors around as a specific type which contains the relevant information separated into different fields,
+rather than combining it all into one single `SDoc`.
+
+Another change that has been made is the introduction of [error codes](https://gitlab.haskell.org/ghc/ghc/-/issues/21684) to uniquely identify each possible error output from GHC. 
+
+These features together serve to make error consumption a breeze for GHC API users, but what about those that would like to consume errors without utilizing GHC's API?
```
michaelpj, line 29: I commented on the issue, but just to repeat here: HLS uses the GHC API, so it's not one of the parties who would benefit from this.

proposals/0000-specify-dump-json.md
```diff
@@ -0,0 +1,155 @@
...
+
+The requirements to solve this problem are to enable the consumption of GHC's diagnostic message in a structured representation that closely matches the internal representation without requiring users to utilize the internals of GHC. The most reasonable solution is to use a compiler flag to dump the structured messages into a versioned JSON output, which is described by a JSON Schema, to standardize the expected output. The hope is that the community can reach a general consensus on the requirements imposed by such a JSON schema. While it will not be too costly to change the JSON schema as new requirements come up (simply increment the schema version), the ideal situation is to present here an opportunity to provide input on the current schema and receive constructive criticism.
+
+## Prior Art and Related Efforts
```
michaelpj, line 41: I think a useful point of comparison is the LSP spec, which defines a type for `Diagnostic`s. While there's certainly no need to mimic what LSP (it has plenty of problems), it might be useful for inspiration.

proposals/0000-specify-dump-json.md
```diff
@@ -0,0 +1,155 @@
...
+ - `"severity"` corresponding to the above `Severity`
+   - required by the schema
+ - `"code"` corresponding to the change in this [proposal](https://github.com/haskellfoundation/tech-proposals/blob/main/proposals/accepted/024-error-messages.md) and the above `DiagnosticCode` in typeclass `Diagnostic`
+   -  this is required by the schema but is allowed to have a value of `null`. The hope is that once all diagnostics officially have error codes, this can be required and nonnullable, but that will happen once `diagnosticCode` itself doesn't return a `DiagnosticCode` wrapped in a `Maybe`
+ - `"hints"` corresponding to the above `[GhcHint]`
+   - The schema itself doesn't validate at all what these may be, only that they are themselves objects. In my view, the best way to handle them is to output a JSON-ified version of the type `GhcHint`. Any attempt to specify them more than that will result in spiraling complexity and a required update every time the `GhcHint` type changes. Another alternative is to just remove them from the output and only present suggestions as they are rendered in the `message`.
```
michaelpj, line 107: Doesn't this rather reduce the usefulness? One example usage of this that people suggested was an "auto-fixer" for issues. Presumably that would run off the hints, and if they're not specified then people are back to ad-hoc string parsing.

michaelpj (commented):
I'd like to see a list of use cases so it was clear that the schema is sufficient for them. See e.g. the question of whether we need structured hints or not.

proposals/0000-specify-dump-json.md
benbellick: That is a good point. In that case, what do you think about presenting a JSONified version of GHC's internal 'GhcHint' type?

michaelpj: I'm not going to be an actual consumer of this so I'm not sure really :shrug: I think part of the issue is that then you presumably will have to either get GHC devs to stabilise the hint type or change the schema frequently.

proposals/0000-specify-dump-json.md
benbellick: I actually liked the way LSP presented spans better, and altered the schema to reflect that. Thanks for sharing!

benbellick:
> I'd like to see a list of use cases so it was clear that the schema is sufficient for them. See e.g. the question of whether we need structured hints or not.

Here is a basic list:
 - A _primary_ use case will be [those who want an LSP-like experience](https://github.com/josephsumabat/static-ls) without using Haskell language server (particularly for projects which may be too large to efficiently benefit from Haskell language server),
 - Another possible use case is mentioned [here](https://gitlab.haskell.org/ghc/ghc/-/issues/19278#note_453663), which is to automatically fix errors based off suggestions, also for large code bases which may struggle to with Haskell language server
 - Additionally, online IDEs, sometimes for educational purposes could leverage the JSON api for easy to use diagnostics displayed directly in the IDE:
   - https://code.world/
   - https://play.haskell.org 

There also very likely will be use cases which we are not foreseeing. That is part of the excitement! 

FYI, I am implementing this as part of [Summer of Haskell](https://summer.haskell.org/). As a consequence, I would like to get started working on this ASAP so that I can aim to finish the bulk of the project by the target date. I propose that the first iteration of this aims to satisfy the bare minimum of requirements to be useful, and then subsequent iterations can add additional features as requests pop up. Therefore, lets just get rid of separate `GhcHint` handling and as a starting baseline, only supply 
 - span of source code causing diagnostic
 - severity of diagnostic
 - the diagnostic code
 - the plain text message output by GHC

It seems everyone agrees (more-or-less) unanimously that all of the above features are absolutely required, and any subsequent features are subject to discussion. Does anyone have a problem pushing the proposal forward just starting with those? If not, I will remove the mention of Hints from the proposal. 

P.S.: I am also happy to keep the discussion going re: `GhcHint` during development, but would like to begin development in the meantime.

goldfirere:
I haven't looked closely here but want to voice strong support for this -- including @benbellick's desire to just jump into coding. Implementing some minimal common ground of features sounds like a great first step. I agree that allowing messages to be structured would be yet a leap forward over this basic step, but the basic step is still really useful without the extra structure.

Just throwing out an idle idea I've had about how to communicate this structure, as an alternative to @david-christiansen's idea of metadata-on-the-side: we could use printf-style formatting codes:

```
"message": "Expected type %t does not match actual type %t in %e";
"types": [ { "string": "Int"; "type": { /* detailed information about Int */ }}, { "string": "Widget"; "type": { /* detailed information about Widget */ }} ];
"expressions": [ { "string": "f x y z"; "expression": { /* detailed information about f x y z */ }} ]
```

The downside of this approach is that it requires consumers to perform substitution even if those consumers don't wish to use the structure. David's approach might be better, but I just wanted to mention this alternative idea.

benbellick:
I have altered the proposal to no longer include the hints in the schema. Instead, the hints are mentioned as an opportunity for future possible work. 

If this looks alright to everyone, is it possible to merge it in so that I can begin working on the implementation? Or is it better for me to begin work now, prior to the proposal's acceptance? This is my first proposal, and I am unsure what is considered typical protocol.

david-christiansen:
For RFC proposals like this one, the acceptance process in the HF's Technical Working Group is only a recommendation by this particular group that a proposal be acted on, based on the refinement and discussion that's occurred. We offer this  venue as a service to the community to seek feedback and discussion, but you don't have to ask us for permission to start working! This is a GHC feature, so you should follow the usual process for contributing to GHC.

I think this will be great, and once you implement it, you'll almost certainly learn things that lead to changes anyway. How about we keep this thread open, as many important stakeholders have already commented here, and then if you feel like you need further feedback, you can post again?

Would you like the TWG to vote as well? We're not the right ones to tell you what your JSON schema should look like - that's more something to work out with prospective users and the rest of the GHC team - but if you want more feedback, I'll ask the rest of the group to chime in.

benbellick:
Thanks everyone! In that case, I'm going to begin work on this initial implementation. I appreciate everyone's thoughtful feedback :)

Ericson2314:
> A primary use case will be [those who want an LSP-like experience](https://github.com/josephsumabat/static-ls) without using Haskell language server (particularly for projects which may be too large to efficiently benefit from Haskell language server)

That project also uses the GHC API.

gbaz:
To summarize our HFTT discussion, we're generally favorable, but all have different takes on what additional structured data may be added in the future. So we generally want to approve as is, but we want noted that as an extensible format, it should be considered to add further structured data on the error message itself in addition, as future users may require.

A few proposals for how to do this have been mooted:
* David suggests spans for annotations: https://github.com/haskellfoundation/tech-proposals/pull/50#issuecomment-1600668883
* Richard suggests printf style annotations: https://github.com/haskellfoundation/tech-proposals/pull/50#issuecomment-1610271254
* John (in committee discussion) suggests a `toJson` of the error message constructors
* I (in committee discussion) suggest a few lists, one for `moduleNames`, one for `typeNames` etc, suitable for editors to do structured highlighting.

gbaz:
(Also, we approved this as an RFC by vote in our meeting, but I'll let David explain in more detail and do the honors).

david-christiansen:
And here I thought you were going to do the honors, @gbaz!

When we approve an RFC, we're saying that this specific group of people think that the right stakeholders have been involved, and that we generally think that the suggestion will improve the state of Haskell. We don't tell you what to do, and you're not bound by any of our suggestions - just in case that needed reiterating.

The committee voted to approve, with five in favor and one abstention (the abstention was due to not having read the discussion in enough detail).

I'm looking forward to this feature - thanks!

silky:
@benbellick Can I recommend you link to this issue, somewhere, in the work you are doing, so that interested parties can follow along :)

benbellick:
Hey everyone, I have an MR open of my work [here](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/11093) if you would like to check it out!

david-christiansen:
Thanks for following up! I'm glad to see this happening.

Are you planning on writing a blog post or similar to describe the feature and the process by which it came about? I think that could be a good way to help draw attention to this new capability.

benbellick:
@david-christiansen Yes I will happily write a blog post once all is said and done and share it on the appropriate channels! Though I am inclined to wait to publish the post once the feature has been merged into GHC.

david-christiansen:
That makes a lot of sense. Let me know if you want feedback on a draft!

flip111:
@benbellick thanks for your effort. The structured message has been talked about in a few comments here, but it's not in [the proposal](https://github.com/benbellick/tech-proposals/blob/specify-json-dump/proposals/0000-specify-dump-json.md) under `For demonstrative purposes, here is an example valid instance of the schema.`. I assume it didn't make it into the proposal.

If the main worry is to rely on complicated GHC types for structure i would suggest to make it available under a different json property alongside the string formatted message.

Sorry to say but i think the proposal is not sufficient as is because you know where a problem is but can not do anything specific with that problem until you parse the message. Since json is mainly useful to be read by other tools the first thing people will want to do is start parsing those messages. In which case i would rather re-use the GHC data structure than re-invent the data structure and write a parser on top. Both the structure as the string are subject to change anyway so tool markers already experience the same amount of effort to keep up with GHC versions.

I have a script where i parse output of the purescript compiler (the cousin language) which has similar error messages. The regexes look like this `%Warning\s+\d+\s+of\s+\d+:\n\n\s+in\s+module\s+([\w\.]+)\n\s+at\s+([\w\./]+):(\d+):(\d+)\s+-\s+(\d+):(\d+)\s+\(line\s+(\d+),\s+column\s+(\d+)\s+-\s+line\s+(\d+),\s+column\s+(\d+)\)\n\n\s+The import of module ([\w\.]+) contains the following unused references:\n\n([(/)\n\s\w]+)\n\n\s+It could be replaced with:\n\n\s*([,\w \.|(|)$&<>=-]+)%`. Just to indicate what it would take to go down the road of string parsing.

benbellick:
Hi @flip111 thanks for the comment! First of all, a valid instance of the schema is indeed presented [here](https://github.com/benbellick/tech-proposals/blob/specify-json-dump/proposals/0000-specify-dump-json.md?plain=1#L114). Or do you mean that the "message" field itself is simply a string? 

I believe what is happening here is confusion between what GHC calls a `Message` and the colloquial use of "message" which is represented by the "message" field in the JSON output presented in the proposal. When I refer to a GHC `Message`, I mean data represented by [these types](https://github.com/benbellick/tech-proposals/blob/specify-json-dump/proposals/0000-specify-dump-json.md?plain=1#L80). The goal of this proposal is **not** to create a structured output for things like "Couldn't match type ‘b0’ with ‘(Bool, b0)’ \n Expected: ...", but rather to provide users with access to the rich information contained inside the `MsgEnvelope` type (e.g., line numbers, error codes, etc.). 

Given that, I believe this proposal does address this issue, as simple JSON parsing provides access to such data. If there is a desire down the line to further breakdown the message itself, that can be worth exploring, but there is no obvious way to standardize that yet. So, this seems like a good start. 

Does that address your concerns? Please let me know if anything is unclear. Thanks!

flip111:
@benbellick  thanks for your answer.

> Or do you mean that the "message" field itself is simply a string?

yes i meant that. [Further up](https://github.com/haskellfoundation/tech-proposals/pull/50/files/f17fbd133a7569441d64af4aaac5a8fd40594937#diff-e592baad12f593a37f160b6d416d3d94e943989e055fc78e6e5c5c5987c10f54R38) you said

> To make sure I am understanding you correctly, you mean that this particular field could be a JSON serialized version of the diagnostic message, rather than the message text itself? That is a great point. I believe that that should be added in addition to the text version.

I understood that you are behind the proposal to create a structured message. But it would be outside of the scope of this particular proposal. Did i understand you correctly? The proposal said this:

> GHC is currently undergoing a long term project to [move to a more structured error representation](https://gitlab.haskell.org/ghc/ghc/-/issues/18516). 
In harmony with this motion, there is

The linked issue talks about decomposing the error message. Perhaps i misunderstood "in harmony with" that this proposal was to address that directly. Instead it should be seen as preparing work to add that later, is that correct?

> If there is a desire down the line to further breakdown the message itself, that can be worth exploring.

When thinking down the line.. I'm wondering whether the implementation without message decomposing can already serve new use cases and create possibilities for tool makers. The original use case was described minimally (only what i could find about it in https://gitlab.haskell.org/ghc/ghc/-/issues/19278 anyway)

> This has the potential to be very useful to e.g. IDEs who want to consume GHC's error messages.

So my question basically boils down to ask about your view, if you rather have a more limited proposal and a follow-up can be done in another proposal, or you would like to pack this together in this same proposal. 

> but there is no obvious way to standardize that yet

I think this is a point of discussion whether the feature should be withhold altogether until it can be standardized. IMO there is benefit to just dump the structured format directly from GHC with the caveat that it will be an unstable format. Because inspecting error messages brings a whole lot new use cases and working on structured data has benefits for tool makers. Also i expect that error messages will be standardize that it will not be difficult to change tooling that already reads the previous unstable error message data structure.

I'm already happy with the smaller scope because it's a good improvement fwiw

david-christiansen:
This is good feedback for follow-up work, @flip111 . The proposal has already been discussed and the work carried out, though, so I don't know that there's much capacity to react to your suggestions right now. Just to make sure you know the status before investing a ton of time in feedback!

PR #47, opened 2022-12-11
Standard library reform: Split base

Ericson2314:
Issues with the standard library are holding back the Haskell ecosystem. The problems and solutions are multifaceted, and so the Haskell Foundation in its "umbrella organization" capacity is uniquely suited to coordinate the fixing of them.

Here is proposed a first step in this process, splitting base.
Future work addressing a larger set of goals is also included to situate splitting base in context.

[Rendered](https://github.com/Ericson2314/tech-proposals/blob/standard-library-reform/proposals/accepted/047-standard-library-reform.rst)

PR #51, opened 2023-06-26
GHC's base libraries: Combining stability with innovation

Ericson2314:
This document describes an plan agreed between the GHC Team and the Core Libraries Committee, saying how we plan to work in partnership to reconcile the goals of innovation and stability for GHC and its ecosystem.

Formally, then, it is not so much a proposal as a record of an outcome. We are, nevertheless, using the Haskell Foundation Technical Working Group proposals repo, so that we can have a permanent public record of how we plan to work, so that others can comment, and so that the document can be polished to add clarity where necessary.

[Rendered](https://github.com/Ericson2314/tech-proposals/blob/ghc-base-libraries/proposals/accepted/051-ghc-base-libraries.rst)


proposals/accepted/051-ghc-base-libraries.rst
```diff
@@ -0,0 +1,330 @@
...
+  - We could consider issuing a warning if you say ``-package ghc-internals`` (or ``ghc-bignum`` or ``ghc-prim``), one that was hard to silence.
+    Since we can have module-level ``WARNING`` pragmas with custom categories, one way to realise this would be to pick a category and add such pragmas to every module in the relevant packages, though we might want to do something more systematic.
+    The text of the warnings could encourage users to
+
+    - switch to a function exposed by base, and/or
+    - petition the CLC to expose this super-useful function from base.
+
+  - ``cabal check`` (a per-package check) could warn on packages that use ``ghc-internals``.
+
+  - ...what else?
```
nomeata, line 128: > We should use all possible social and technical means to discourage people from depending directly on ``ghc-internals``

Really all? Including GHC/Cabal simply refusing to allow packages called  `ghc-internals` if the package currently built isn't `base` or another internal whitelisted package?

But it says “discourage”, not “prevent”, which sounds reasonable.


Maybe more realistic: hackage could reject uploads of packages that depend on `ghc-internals`?

Ericson2314: > But it says “discourage”, not “prevent”, which sounds reasonable.

Yes that is key. If people want to tinker with the internals of some open source software, we should not stop them. But we should make sure they are very aware they are stepping outside the stable safe zone. And likewise we ought to someday make anyone that uses the tinker's software (transitively!) also aware they are exposed to instability.

proposals/accepted/051-ghc-base-libraries.rst
```diff
@@ -0,0 +1,330 @@
...
+  (But CLC would not expect to curate the API of ``ghc-experimental``.)
+
+- Perhaps ``ghc-experimental`` should be in the purview of the GHC Proposals process.
+  GHC devs should not just make up random APIs and pop them into ``ghc-experimental``; a scrutiny process would be valuable.
+
+- Under this proposal, there is initially no change (whatsoever) to the API exposed by ``base``, or its performance characteristics.
+  The impact on clients should therefore be zero.
+
+  Over time, the GHC developers may make CLC proposals to remove types and functions that are currently in the ``base`` API, but are in truth part of GHC's implementation, and were originally exposed by historical accident.
+  But these are *future*\ proposals.
```
mixphix, line 146:
```suggestion
  But these are *future* proposals.
```

proposals/accepted/051-ghc-base-libraries.rst
```diff
@@ -0,0 +1,330 @@
...
+
+  To make the transition suggested in these future proposals easier to manage, we have in progress a "deprecated exports" mechanism that will ease such transitions.
```
mixphix, line 148:
```suggestion
  To make the transition suggested in these future proposals easier to manage, we have in progress a ["deprecated exports"](https://github.com/ghc-proposals/ghc-proposals/pull/595) mechanism that will ease such transitions.
```

proposals/accepted/051-ghc-base-libraries.rst
```diff
@@ -0,0 +1,330 @@
...
+  For a transitional period, ``base`` can continue to export the function, but with a deprecation warning saying something like:
+
+    This is going to disappear from base.
+    You probably don't want to use it at all.
+    But if you absolutely must, get it from ``ghc-internals``.
```
mixphix, line 153: We can't say in the same document that we should use "all means necessary" to avoid `ghc-internals` dependencies, and also recommend people to import a function from there.

```suggestion
    If you absolutely must, define it yourself locally, or if this is impossible, let the CLC and GHC team know!
```

proposals/accepted/051-ghc-base-libraries.rst
```diff
@@ -0,0 +1,330 @@
...
+    Documentation changes normally fall under GHC developers purview, except significant ones (e.g., adding or changing type class laws).
+
+- It also supports GHC innovation, by
+
+  - allowing GHC freedom to change aspects of its implementation
+
+  - allowing the GHC Steering Committee to add new functions and types in ``ghc-experimental``.
+
+- One might wonder why GHC has three "internal" packages: ``ghc-internals``, ``ghc-bignum``, and ``ghc-prim``? Could they not be a single package? Answer: technically yes, but it helps to keep dependencies and responsibilities clear.
+  And it's purely an internal GHC matter; if the team wants to structure GHC's internals with three packages, or ten, that's up to them.
```
mixphix, line 173: Let's change this awkward phrasing:

```suggestion
- The three "internal" packages: ``ghc-internals``, ``ghc-bignum``, and ``ghc-prim``, could arguably be a single package, but the GHC team has decided that encapsulating the relevant code in this way helps to keep dependencies and responsibilities clear.
```

mixphix (changes requested):
Just a few minor suggestions, and one point worth noting: "all possible technical and social means to discourage" means not actively encouraging something ourselves. Say something and stick to it!

I rather like @nomeata's suggestion that Hackage reject packages with explicit `ghc-internals` dependencies, it seems like the right abstraction layer to do so: if someone *really, really* wants to depend on it, they can still do so privately.

proposals/accepted/051-ghc-base-libraries.rst
Ericson2314: I am fond of the second sentence, because its perhapps the only indication that `ghc-internals`'s very existence is also unstable; this follows from the GHC devs being allows to shuffle around code amongst their unstable libraries as often as they please.

proposals/accepted/051-ghc-base-libraries.rst
Ericson2314: I would expect `ghc-internals` to mainly contain things one cannot define locally? I would expect also some zany integrate-with-ghc-so-tightly-someone-turns-blue projects to always need to use `ghc-internals` too. I think it's less important to deny that category exists, than emphasize that the vast majority of projects are *not* in that category.

Put another way, I'd expect those projects to know who they are and not reach out to GHC devs or the CLC, and I'd expect almost everyone who does reach out to be told "actually, you don't need to do things that way". Is that what you were thinking too? Or were you imagining we might sometimes move something to `ghc-internals` (without a reexport elsewhere) that shouldn't be there, by mistake?

mixphix: If such projects are relying on functionality from `base` that GHC developers and the CLC both agree should be moved to `ghc-internals`, they ought to reach out and say "hey, we use this for such-and-such", and make a case for it to be put in `ghc-prim` instead.

The point here is to make, FAIAP, a "hidden" library containing code in the strictest intersection of "compiler bootstrap" and "necessary to allow general-purpose programming": stuff like the guts of IO, datatype representations, and the like. There should be nothing  in `ghc-internals` that is both:
- useful for general programming, and
- not exported from `base`, `ghc-prim`, or `ghc-experimental`.

Ericson2314: I agree with the main "There should be nothing in ghc-internals that is both: useful for general programming, and not exported from..." point, but do not that I don't think anyone intends `ghc-prim` to be any more user-facing than `ghc-internals`. Perhaps the proposal should make this more explicit?

hasufell:
> I rather like @nomeata's suggestion that Hackage reject packages with explicit ghc-internals dependencies, it seems like the right abstraction layer to do so: if someone really, really wants to depend on it, they can still do so privately.

I disagree.

ghc-internals is internal API, but properly PVP versioned. Power users writing high performance code or **alternative Preludes** may have very legitimate use cases of importing it. Rejecting those packages on hackage is wrong.

proposals/accepted/051-ghc-base-libraries.rst
Ericson2314: (The difference between `ghc-prim` and `ghc-internal` also isn't clear. But one answer is

> GHC devs can structure (and restructure) their internals however they like, including dividing definitions *arbitrarily* between `ghc-prim` and `ghc-internals` if they so desire, so this question doesn't need to be answered in this document.

Another answer is

> `ghc-prim` is (mostly, perhaps should be entirely) auto-generated docs for code that doesn't actually exist in textual form (it's just hard-coded in GHC), whereas `ghc-internals` is human-written code.

)

nomeata:
> Power users writing high performance code or alternative Preludes may have very legitimate use cases of importing it. 

That sounds very reasonable, but very different from “discourage by all means possible”, and probably deserves clarification. Which one is it?

hasufell:
> > Power users writing high performance code or alternative Preludes may have very legitimate use cases of importing it. 
> 
> That sounds very reasonable, but very different from “discourage by all means possible”, and probably deserves clarification. Which one is it?

"Discourage" is the correct wording, it does not imply "prohibit".

angerman:
:+1: This seems like a step in the right direction. I hope that this is the first step of decoupling GHC and GHC releases from library upgrades rippling through the whole ecosystem. 🤞

angerman (commented):
I will leave a few comments:
- The title is: "GHC's base libraries: Combining stability with innovation". Yet, it doesn't define any of those precisely. It lists `base`, `ghc-prim` somewhere in the documents. Then `ghc-internals`, `ghc-bignum`. We may want to call out what exactly we understand by "base librarie**s**".
- What is understood by "stability"?
- What is understood by "innovation"?

> The two goals are in risk of conflict, and we have seen such conflict in the past.

Feels like it unnecessarily pits the CLC and GHC team against each other. As someone who has contributed to GHC, and had whose primary interest is stability, and proper system integration, it feels unnecessarily discouraging. We do want people to contribute to GHC, that care about stability.

I strongly disagree that we want to have a CLC (stability stalwarts) vs. GHC (innovators) view set out here. GHC needs to welcome people who care about stability and backwards compatibility and the CLC also needs to welcome advocates for innovation.

hasufell:
> > The two goals are in risk of conflict, and we have seen such conflict in the past.
> 
> Feels like it unnecessarily pits the CLC and GHC team against each other. As someone who has contributed to GHC, and had whose primary interest is stability, and proper system integration, it feels unnecessarily discouraging. We do want people to contribute to GHC, that care about stability.
> 
> I strongly disagree that we want to have a CLC (stability stalwarts) vs. GHC (innovators) view set out here. GHC needs to welcome people who care about stability and backwards compatibility and the CLC also needs to welcome advocates for innovation.

This is why I had proposed a different wording that doesn't focus on exact goals, but on the autonomy of GHC and CLC, which makes sense for its own sake:

----

1. CLC has full autonomy over decisions affecting 'base' (API, performance, semantics), including its dependencies.
2. GHC developers seek more autonomy over GHC internals and experimental features.

We want CLC and GHC to be able to exercise different goals (e.g. stability) and workflows (release management/versioning), but be sure that CLC's decision-making is not compromised, e.g. by changes that indirectly affect 'base'. We also want to be efficient and avoid proposals that are not strictly relevant to proper 'base' API.

----

@simonpj was unhappy with "including its dependencies", which I still don't understand, because it's absolutely true.

So let me propose this alternative wording again. Let's not focus on exact goals, because they are rather hard to express when two heterogeneous groups are involved. Let's focus on autonomy and how collaboration is supposed to work.

chreekat:
A quick comment on the question of hiding `ghc-internals` on Hackage. My experience searching for information that has been "helpfully" hidden from me leads me to recommend against. Much better would be to build a standard way to mark a library as "internal" that tooling can use to put big honking warnings around. More information rather than less information. Plus it would allow other library authors to use the same mechanism, ameliorating a general problem and playing well with the support for multiple public libraries that's shaping up.

A caveat: for now, the "big honking warnings" should just go in the Haddocks, to decouple this effort from *any* hypothetical tooling changes.

proposals/accepted/051-ghc-base-libraries.rst
```diff
@@ -0,0 +1,330 @@
...
+
+   - Move rapidly to fix bugs, improve performance, and refactor GHC's internals to pay down technical debt.
+
+The two goals are in risk of conflict, and we have seen such conflict in the past.
+The purpose of this proposal is to reconcile these goals, in a way that satisfies all parties.
+
+Each of us places different emphasis on these goals, but respects them both.
+The CLC does not want to constrain GHC unnecessarily; and the GHC team does not want to undermine CLC's efforts, indeed quite the reverse!
+
+The proposal is based on `HF Proposal: standard library reform <https://github.com/haskellfoundation/tech-proposals/pull/47>`__, but is independent of it.
```
simonpj, line 50: Add: 

You can find more background in [CLC issue #146](https://github.com/haskell/core-libraries-committee/issues/146).

Ericson2314: Like this?
```suggestion
The proposal is based on `HF Proposal: standard library reform <https://github.com/haskellfoundation/tech-proposals/pull/47>`__, but is independent of it.
You can find more background in `CLC issue #146 <https://github.com/haskell/core-libraries-committee/issues/146>`__.
```

simonpj: Yes, that's what I was thinking.  Just to keep pointers to history.

proposals/accepted/051-ghc-base-libraries.rst
simonpj: > not exported from base, ghc-prim, or ghc-experimental.

This should be "not exported fomr base or ghc-experimental".  `ghc-prim` is every bit "internal" as `ghc-internals`.   Anything in `ghc-prim` that is useful for general programming should be be exposed by `base` or `ghc-experimenal`.  Indeed maybe we should make that clearer.

proposals/accepted/051-ghc-base-libraries.rst
simonpj: @mixphix's rephrasing of the first sentence is helpful. (I would remove "arguably" or replace with "in principle".)  But as John says, let's retain "It's purely an internal GHC matter"

simonpj:
> So let me propose this alternative wording again. Let's not focus on exact goals, because they are rather hard to express when two heterogeneous groups are involved. Let's focus on autonomy and how collaboration is supposed to work.

Indeed, the wording in the document is very close indeed to what you suggest (not a coincidence) including "including its dependencies".   (I elaborated a bit what the GHC team seeks.)   Perhaps there are other changes you would like to see?

>  GHC needs to welcome people who care about stability and backwards compatibility and the CLC also needs to welcome advocates for innovation.

I strongly agree with this sentiment @angerman!  Indeed the doc says "Each of us places different emphasis on these goals, but respects them both. The CLC does not want to constrain GHC unnecessarily; and the GHC team does not want to undermine CLC's efforts, indeed quite the reverse!".   So I think we all agree about what we want to say; the only question is, how to say it.  Do you have a concrete suggestion for how we might say this better?  Clearly it is not coming over as intended, to you anyway.  Let's fix that!

proposals/accepted/051-ghc-base-libraries.rst
```diff
@@ -0,0 +1,330 @@
...
+  Julian Ospald — CLC,
+  Simon Peyton Jones — GHC
+
+.. sectnum::
+.. contents::
+
+Introduction
+=========
+
+This document describes an plan agreed between the GHC Team and the Core Libraries Committee, saying how we plan to work in partnership to reconcile the goals of innovation and stability for GHC and its ecosystem.
```
ocharles, line 23:
```suggestion
This document describes a plan agreed between the GHC Team and the Core Libraries Committee, saying how we plan to work in partnership to reconcile the goals of innovation and stability for GHC and its ecosystem.
```

angerman:
@simonpj I tried rewording it a few times. It's hard. Harder than I thought it would be. I actually like @hasufell's wording a lot. The current wording in the proposal seems to lead too strong with "The CLC" vs "The GHC Team", and assigning them different roles from the the get-go in the itemisation. The current phrasing

> - The Core Libraries Committee seeks to exercise its control over decisions affecting 'base' (API, performance, semantics), including its dependencies.
> - The GHC team seeks the freedom to

make it sound incompatible from the onset. Which I don't think it is, nor needs to be.

Contrary to that I feel that @hasufell's wording is much more neutral:
> - CLC has full autonomy over decisions affecting 'base' (API, performance, semantics), including its dependencies.
> - GHC developers seek more autonomy over GHC internals and experimental features.

While I find 
> We want CLC and GHC to be able to exercise different goals (e.g. stability) and workflows (release management/versioning), but be sure that CLC's decision-making is not compromised, e.g. by changes that indirectly affect 'base'. We also want to be efficient and avoid proposals that are not strictly relevant to proper 'base' API.

to come across bit one-sided. We should probably add:

> At the same time we do not want to constrain the development of GHC itself.

Ultimately I see the base split as

```
-----------------------.
                       |
ghc - ghc-internals - base - <wider ecosystem>
                       |
-----------------------'                                       
```

`base` as well as the other libraries mostly under supervision of the CLC are the _public interface_ of the standard library exposed to the community. And we want to ensure that the wider ecosystem can seamlessly depend on a stable interface, while GHC has a bit more freedom to change (and the acknowledgement that base is a bit too fat for this right now).

If we can keep the public interface stable, we can trivially keep compatibility across a few versions, which to me sounds like one of the goal of this proposal. Maybe I'm mistaken?

proposals/accepted/051-ghc-base-libraries.rst
```diff
@@ -0,0 +1,330 @@
...
+
+We propose to divide ``base`` into three packages:
+
+- ``ghc-internals``: exposes aspects of GHC's internals that may be of interest to "hard-core" developers interested in maximum performance (see `Nikita's blog post <https://nikita-volkov.github.io/internal-convention-is-a-mistake/>`__).
+  The API of ``ghc-internals`` is fully under the control of the GHC team, and of no direct interest to the CLC — only its effects on the API of base.
+
+- ``base``: as now, whose API is curated by CLC.
+  Depends on ``ghc-internals``, and hence on ``ghc-bignum`` and ``ghc-prim``.
+
+- ``ghc-experimental``, initially empty, depends on ``base`` and on ``ghc-internals``.
```
parsonsmatt, line 92: OK, so the governance structure is:

- `base` governed by CLC and CLC proposals
- `ghc-internals` governed solely by GHC team, no proposal necessary
- `ghc-experimental` governed by GHC proposal process

The Win here is that GHC Team can unilaterally make changes to `ghc-internals` without asking anyone, which allows them to iterate quickly. GHC Proposals can go in `ghc-experimental` without involving CLC directly. And CLC has less extraneous work. Feels like a win-win-win to me.

proposals/accepted/051-ghc-base-libraries.rst
```diff
@@ -0,0 +1,330 @@
...
+  Functions and data types here are intended to have their ultimate home in base, but while they are settling down they are subject to much weaker stability guarantees.
```
parsonsmatt, line 93: Something I'm always wary of here is the tendency for things to go stale - not quite popular enough to move to `base`, but useful enough to stay in `ghc-{experimental,internal}`. Is there room for a policy on how long something should be "unstable" before getting shelved or promoted?

proposals/accepted/051-ghc-base-libraries.rst
```diff
@@ -0,0 +1,330 @@
...
+
+  As its name suggests, the API of ``ghc-experimental`` is curated by the GHC team, although the CLC is willing to offer (non-binding) opinions, if consulted.
+
+All three packages conform rigorously to the PVP.
+(But see Section 5.3)
+
+Some observations about this structure:
+
+- We should use all possible social and technical means to discourage people from depending directly on ``ghc-internals``, because if such dependencies become frequent and ossified, it will lead to future pain when the API changes.
+  Saying "we told you not to rely on it" may be true, but won't lessen that pain.
```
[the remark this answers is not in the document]
parsonsmatt: I also don't feel like this part of the proposal is well-motivated - specifically why `ghc-internals` should be treated separately from a potential `text-internals` or `bytestring-internals`. 

From my perspective, writing `build-depends: ghc-internals` is just as prone to pain as any other unbounded dependency. And `build-depends: ghc-internals == 9.6.4.*` should be perfectly safe, unless PVP is violated. But the *point* of this proposal is that we can make "breaking changes" to `ghc-internals` as a major version bump without incurring a major version bump for `base` - in other words, so we *can* be PVP compliant.

proposals/accepted/051-ghc-base-libraries.rst
```diff
@@ -0,0 +1,330 @@
...
+And mistakes can lead to a loss of trust.
+
+The solution is obvious: we need to automate.
+We therefore propose the following, as part of CI:
+
+1. Compile a good chunk of Hackage (around 500 packages) against base.
+   We already do this, and it is a huge help in reassuring ourselves that a change does not lead to accidental breakage.
+
+2. Test if any of the types (incl their kinds), functions (incl their types) and instances exposed by the ``base`` API are accidentally changed by a commit.
+   This is definitely going to happen, soon: @bgamari already has a prototype.
```
parsonsmatt, line 190: This tool would be *amazing* to have for other libraries - I'd love to have something here so I could know for sure if a new release is versioned appropriately.

proposals/accepted/051-ghc-base-libraries.rst
```diff
@@ -0,0 +1,330 @@
...
+    Ambiguous occurrence ‘id’
+               It could refer to either ‘Prelude.id’,
+                            imported from ‘Prelude’ at mod153.hs:2:8
+                            (and originally defined in ‘GHC.Base’)
+                         or ‘M.id’, defined at mod153.hs:2:21
+
+  The "originally defined in" mentions a module; and if that module is in a package that is not imported, GHC will package-qualify the module name.
+  And seeing ``ghc-internals:GHC.Base`` is perhaps less nice.
+  This is not a new problem: we already package-qualify modules in ``ghc-prim``.
+  One solution is to remove the "originally defined in.." parenthesis for types and functions that would require such package qualification.
```
parsonsmatt, line 260: IMO a lot of this would be helped by tracking the import/re-export provenance more directly. "Imported from A and originally defined in E" is nice but it's also often very nice to know the exact chain of imports and re-exports that brought something into scope. GHC could then have a flag to trim this provenance information to the direct dependencies of a package. That way, if you find yourself depending on `ghc-internals:SecretType`, you can see exactly how it got introduced into your codebase.

proposals/accepted/051-ghc-base-libraries.rst
```diff
@@ -0,0 +1,330 @@
...
+
+- Another form of leakage could be: a new class in ``ghc-internals``, *not exposed in base*, that is given instances for existing data types.
+  There is a risk that those instances might confusingly be visible to clients of ``base``.
+  If so, the CLC should at least be consulted.
```
parsonsmatt, line 264: Wouldn't you need to import the relevant class to have these instances be visible? And if it's only exposed in `ghc-internals`, wouldn't that require an explicit import of the class?

This may just be an artifact of how [instance visibility works across component boundaries](https://gitlab.haskell.org/ghc/ghc/-/issues/20529), though.

proposals/accepted/051-ghc-base-libraries.rst
```diff
@@ -0,0 +1,330 @@
...
+Very occasionally a **bug-fix** might involve a change to the user-visible API.
+Example: `role annotations on SNat <https://github.com/haskell/core-libraries-committee/issues/170>`__ (although there is a debate as to whether this specific change constitutes a "breaking change" under the PVP).
+
+Under these circumstances we (together) will have to decide whether to
+
+- Back port the fix, and not bump the major version of ``base`` (i.e. bend the PVP), or
+- Bump the major version of base, but therefore be unable to fix the bug in the released GHC.
+
+This is a decision for the CLC.
+See PVP issue https://github.com/haskell/pvp/issues/10.
```
parsonsmatt, line 290: This is a technical limitation of the way that `base` and GHC work, and would in principle be fixed by allow a GHC version to work with multiple versions of `base`, right?

parsonsmatt (commented):
The core part of the proposal feels great.

proposals/accepted/051-ghc-base-libraries.rst
ocharles: I think this already exists as https://kowainik.github.io/posts/policeman-bristol - or is what you're looking for something else?

proposals/accepted/051-ghc-base-libraries.rst
Ericson2314:
```suggestion
- The three "internal" packages: ``ghc-internals``, ``ghc-bignum``, and ``ghc-prim``, could arguably be a single package, but the GHC team has decided that encapsulating the relevant code in this way helps to keep dependencies and responsibilities clear.
  And it's purely an internal GHC matter; if the team wants to structure GHC's internals with three packages, or ten, that's up to them.
```

proposals/accepted/051-ghc-base-libraries.rst
Ericson2314: In asking https://github.com/ghc-proposals/ghc-proposals/pull/595#issuecomment-1607844684 I was wondering something similar --- ideally we we do now know the entire provenance of imports and (re)exports, and there are a number of things we can do with it.

proposals/accepted/051-ghc-base-libraries.rst
Ericson2314: Are you proposing (eventually) to continue to ship the `base` with the bug and a new one with the bug? That would work. But I think the point in this case was that some bugs are not a matter of implementation but inherent to the interface itself.

(I remain very pro multiple `base` versions with one GHC, to be clear :).)

parsonsmatt: I think I'm confused - can a released GHC use a different version of `base`? I was under the impression that the `base <-> GHC` relationship was fixed, and every version of GHC can only use a single version of `base`. That means that GHC X.Y.Z must always use the same `base` version, and a new `base` version means a new release of GHC.

Ericson2314: I think there is much confusion around this topic because people confuse GHC the software with GHC the official bindist.

The answer is Yes, and it's not even a new feature. See https://gitlab.haskell.org/ghc/ghc/-/blob/master/compiler/GHC/Unit/Types.hs#L562.

The short version that is one is liable to shoot themselves in the foot if they have more than one `base` in a package database, but one is always free to have multiple package databases they will not mix together. In each "parallel universe" of the separate package database, there is only one `base`, and no potential for confusion.

GHC in fact has *no way of knowing* whether this is the "official base" that is was shipped with the bindist, or ones own. How could it now? (It isn't shipped with trusted ABI hashes for `base`, for example, on the contrary, it expects `base` to have *no* ABI hashes; it is maximally trusting of whatever `base` you give it!)

For a variety of reasons, no one has yet proposed we take advantage of this. But we can. There is no technical limitation, just the ergonomic/human problem of it being easy to get our package databases with their separate `base`s mixed up.

(I believe @alt-romes might be working on getting all wired-ing packages ABI hashes to solve the humans-getting-confused problem, but I am not sure about that.)

simonpj: > (I remain very pro multiple base versions with one GHC, to be clear :).)

I think we all are in favour of that.   

But, to be clear, achieving that goal is not part of the current document.  So this thread is somewhat "by the way".

Still, the plan outlined this document should make the goal of a reinstallable base much more achievable.  GHC knows (wired into its binary) where, say `GHC.Base.map` is defined.  If we move that definition from one module to another, everything will stop working.  I think that is one reason that `base` and GHC are so tightly bound together.  But if  `GHC.Base.map` was in `ghc-internals` then GHC and `ghc-internals` would remain tightly bound together, but `base` would be much more loosely coupled.  I don't yet see any technical obstacle to installing a fresh `base` without changing `ghc`.

I could be wrong; but all that matters for *this* conversation is that the plan here is a step in the right direction for the reinstallable-base goal.

proposals/accepted/051-ghc-base-libraries.rst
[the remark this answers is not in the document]
simonpj: > I also don't feel like this part of the proposal is well-motivated - specifically why ghc-internals should be treated separately from a potential text-internals or bytestring-internals.

It's really a response to a legitimate CLC concern that if it's too easy to depend on `ghc-internals` then that's what people will do.  And if that happens we might grow an ecosystem of libraries many of whch depend, perhaps by accident, on `ghc-internals`.  So every new release of GHC would force updates to all of those libraries.  You could say "if you depend on `ghc-internals` then that's what you get, but the pain is real and not mitigated by saying "I told you so".  To put it another way, it risks entirely bypassing the stability efforts of the CLC.

So that is, I think, the reason for this text. You are right that all the same issues apply to any `-internals` package, and one might wonder about mechanisms to discourage depending on them. But to avoid scope creep in this document we just stuck to `ghc-internals`.

simonpj:
> @simonpj I tried rewording it a few times. It's hard. Harder than I thought it would be. I actually like @hasufell's wording a lot. The current wording in the proposal seems to lead too strong with "The CLC" vs "The GHC Team", and assigning them different roles from the the get-go in the itemisation.

Thanks.  I'm not getting the "setting up conflict" vibes, but I'm entirely happy to change the words, since I think we all agree on the intent.  How about this?

-------------
This proposal allows the GHC team and the Core Libraries Committee to work in a productive partnership.  What we want to achieve is this:

* CLC has full autonomy over decisions affecting `base` (API, performance, semantics), including the indirect consequences on `base` of changes to its dependencies.
* The GHC team has maximum freedom (consistent with the CLC's curation of `base`) to:
  * Innovate in the language design. GHC has hundreds of extensions, and people suggest more all the time, via the GHC Proposals process.
  * Move rapidly to fix bugs, improve performance, and refactor GHC's internals to pay down technical debt.

The proposal sets up mechanisms (e.g. what packages exist), responsibilities (e.g. who curates which package), and processes (e.g. who should be consulted and when) that support both CLC and the GHC team to follow their respective goals without tripping over each other.

The proposal is based on https://github.com/haskellfoundation/tech-proposals/pull/47, but is independent of it. You can find more background in https://github.com/haskell/core-libraries-committee/issues/146.

proposals/accepted/051-ghc-base-libraries.rst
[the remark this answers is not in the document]
hasufell: Discouragement is well motivated. Prohibition isn't.

We have to educate users more generally about the pattern we're trying to introduce here and how it relates to PVP, API stability, etc.

I could imagine that we try to summarize the result of these efforts for end users and library maintainers in a more distilled way and explain the pattern there in more depth.

I don't think we have to be particularly scared about ecosystem degrading.

1. Hackage has a reverse deps feature and we can monitor packages depending on ghc-internals easily.
2. There are many eyes on popular packages like aeson, servant, conduit etc. and I'm sure someone will scold any attempt at introducing a ghc-internals dependency
3. Even if a package accidentally does so, it's not irreversible. We can scold the maintainer and ask why they saw the need to do so.
4. "Pick your libraries carefully" already applies. Well educated maintainers who care about the ecosystem will play nicely.

proposals/accepted/051-ghc-base-libraries.rst
bgamari: `policeman` is a good approximation but when I examined it for use in GHC I found it had a few shortcomings:

 * it makes no attempt at comparing anything beyond the names of exports; that is, the types of bindings, exposed instances, `MINIMAL` declarations of classes,  and the like are not accounted for in the PVP assessment
 * there isn't a clear way to handle packages with platform-dependent exports; `base` is (unfortunately) such a package
 * it relies on `.hie` files, which we currently don't produce in the GHC build

proposals/accepted/051-ghc-base-libraries.rst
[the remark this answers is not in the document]
parsonsmatt: @simonpj I think that makes good sense. I want GHC to be able to innovate freely and iterate quickly, and I wouldn't want CLC to start encroaching on the `ghc-internals` package for this reason. Where I think I'd draw the line is at describing the problem, noting it is a known ecosystem-wide problem that deserves a good solution, and somehow put a reminder on the package to *use* that solution when it is implemented.

There are a few perspectives here, and I think it'd be good to be really clear about which perspective a bullet is trying to satisfy. This one appears to be trying to satisfy both "industrial users want fewer breaking changes when upgrading major versions of GHC" and "CLC wants to be able to promote stability for the ecosystem" - which are overlapping but subtly different concerns.

As an industry user and OSS contributor, the primary pain point of upgrading GHC is that there are breaking changes that require a large chunk of the ecosystem to be touched. `base` is usually a relatively small part of this pain, with `ghc-prim` and `template-haskell` causing far more issues. I don't expect that `ghc-internals` is going to meaningfully impact the work that gets done here.

If `ghc-internals` is banned by `cabal-install`, then I don't see how it is different from un-exposed modules in `base`. If `ghc-internals` is banned on Hackage, then industrial users will need to vendor code in order to share libraries that depend on it - which *increases* the pain of a major upgrade by requiring *applications* to depend on `ghc-internals` directly, rather than sharing a library which can abstract over multiple versions of `ghc-internals` via CPP or other techniques. If there are warnings, they will be ignored for good and bad reasons, and maintainers will have another package to consider when doing ecosystem upgrades, but the overall workload doesn't change much.

As a CLC member, `ghc-internals` makes my life easier by having fewer changes to `base`, and changes to `base` that can be backed up by real ecosystem use in `ghc-internals`. But if using `ghc-internals` is heavily discouraged, then it's difficult to identify whether or not something in that package is satisfying the needs of the ecosystem.

Summarizing,

- I don't think we should use strong wording like "all possible social and technical means." 
- I'm not even sure we need to discourage beyond calling the package `ghc-internals` and writing docs that it'll be a major version bump with breaking changes every GHC release.
- I *definitely* think that proposing specific technical means for discouraging folks is out-of-scope here.

proposals/accepted/051-ghc-base-libraries.rst
parsonsmatt: Fantastic, thanks for the clarification!

Allowing `base` to iterate without incurring a GHC release cycle would be a big win for the ecosystem, for sure.

proposals/accepted/051-ghc-base-libraries.rst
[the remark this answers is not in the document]
gbaz: The more I think about it, the more I am convinced that "all possible social and technical means" is too strong. It is obviously not literal. We are not going to hire a crack commando unit who survive as soldiers of fortune to go to your house if you depend on it.

Better language might be "develop both social and technical mechanisms to discourage..."

proposals/accepted/051-ghc-base-libraries.rst
bgamari: Not necessarily. The primary way that this exposure might happen is via Haddocks. For instance, imagine that we have a class, exposed via `ghc-experimental` (and not in `base`) but defined in `ghc-internals` which some class exposed via `base` is an instance of. Under `haddock`'s current logic, this instance would be shown in the documentation for `base`, despite the class itself not being visible in `base`.

michaelpj:
I find the lifecycle of and responsibility for `ghc-experimental` a bit weird.

- It is a staging ground for things going into `base`
  - ... but it has `ghc` in the name
  - ... but it is the responsibility of and gatekept by the GHC Steering Committee. 
- If it is valuable to have such a staging ground, why restrict it to just changes that come from GHC proposals?
- You need the approval of two bodies now: the GHC Steering committee to get into `ghc-experimental` and then then the CLC to get into `base`. Splitting these approvals seems both counterintuitive and potentially a source of trouble. 
    - It seems like the CLC could refuse to adopt something from `ghc-experimental`, which would leave things in a bad state (GHC proposal accepted, but code stuck in `ghc-experimental`). Or are they expected not to do this? Why not?
    - As a contributor, if the CLC must eventually approve my changes, I would very much like them to weigh in early before I do lots of work!

Here's an alternative proposal:

- The library is called `base-experimental`
- `base-experimental` is the responsibility of and gatekept by the CLC
- In order to get something into `base-experimental` you need to get an agreement-in-principle from the CLC that it would be acceptable in `base`. That could mean "yes, once it's stabilized", or "yes, once you've addressed specific problem X that seems solvable", or "yes once we've bikeshedded the naming", or similar
- Moving from `base-experimental` to `base` requires the CLC to approve the final version. The CLC's response should almost always be "yes if", and only exceptionally "no"
- GHC proposals can propose either to add something to `base` or `base-experimental`. In either case the corresponding level of approval (agreement with the precise proposal for `base`, agreement-in-principle for `base-experimental`) from the CLC is required for the proposal to be accepted.

I think this would have some advantages:

- The CLC is clearly responsible for everything `base`-related, the GHC Steering Committee is not responsible for `base` at all
- `base-experimental` is a generic place where good-but-unrefined additions to `base` can stabilize, and is not restricted to GHC proposals
- The lifecycle of things in `base-experimental` is clear

The downside is that the CLC has to get involved earlier. But I think the current proposal is false economy: since the CLC will have to get involved later _anyway_, there doesn't seem to me to be much advantage in delaying it.

proposals/accepted/051-ghc-base-libraries.rst
michaelpj: 100% agree that `ghc-experimental` should have a clear lifecycle, and ideally an "up-or-out" policy of some kind.

proposals/accepted/051-ghc-base-libraries.rst
```diff
@@ -0,0 +1,331 @@
...
+
+  - Cabal description and README explains how it is intended used (and not used).
+
+  - Hoogle could (by default anyway) never show stuff from ``ghc-internals``.
+
+  - Do not upload Haddocks for ``ghc-internals`` to Hackage.
+    (Ditto ``ghc-prim``.) Need to make sure that if someone wants to follow the Haddock source-code link to (say) Functor, they should still find it regardless of where it is actually defined.
+
+  - We could consider issuing a warning if you say ``-package ghc-internals`` (or ``ghc-bignum`` or ``ghc-prim``), one that was hard to silence.
+    Since we can have module-level ``WARNING`` pragmas with custom categories, one way to realise this would be to pick a category and add such pragmas to every module in the relevant packages, though we might want to do something more systematic.
```
michaelpj, line 121: This sounds excellent. Much like "partial", "internal" seems like a fairly natural warning category that other people in the ecosystem might want to use. And it's easy for a user to say "I know what I'm doing" by just turning off that warning category in the module where they use internals.

tomjaguarpaw:
> As a contributor, if the CLC must eventually approve my changes, I would very much like them to weigh in early before I do lots of work!

I don't understand this objection. If you want something in GHC then you propose it to Team GHC and, if successful, you and/or they add it, and parts of it may be exposed through `ghc-experimental`. You're then more than welcome to expose a stable interface to your work through your own package. It really has nothing to do with `base` at all!

tomjaguarpaw:
> If I understand you rightly, you would say "This is fine, in such a case the author can just distribute that code via another package"?

Yes, absolutely. If and when appropriate (based on usage numbers, stability, and other concerns) the functionality can be absorbed into `base`. This has been my expectation of how the `ghc-experimental` flow could work since I first heard of it.

simonpj:
> Can we add some examples of the various workflows to the proposal, then? I think that would help a lot.

If anyone would like to offer to some concrete text, that would be great.  We could have a section at the end on "possible workflows".   

I'm listening to @bodigrim who counsels against scope creeep.  But a clearly-seprarated discussion/illustration section, that doesn't form part of the proposal but has dicussion about how it might work in practice, would be fine I think.  It just needs someone to write it!

proposals/accepted/051-ghc-base-libraries.rst
gbaz: There's the `category` field in cabal files which is pretty unstructured an all over the map. We could easily initiate a convention of the `internal` category actually meaning something -- e.g. triggering an extra textual notice on hackage pages listing the package, and extending cabal so it could maybe optionally filter or warn on such stuff.

hasufell:
> I find the lifecycle of and responsibility for `ghc-experimental` a bit weird.
> 
> - It is a staging ground for things going into `base`

No, it is not.

`ghc-experimental` is not under CLC purview and not maintained by CLC. It's entirely up to GHC HQ/SC how to use it.

CLC doesn't necessarily care about it at all. But it's a possible mechanism for GHC to:

- test and expose new API early
- demonstrate the experimental API to CLC when doing a proposal

Proposals may still be rejected. We have full autonomy. 

GHC SC is welcome (and encouraged) to source CLC's opinion before accepting GHC proposals that possibly touch base, but it's not a requirement.

But if you don't, you can't complain later that your proposal as a whole failed due to CLC rejecting it.

tomjaguarpaw:
> Is there a picture anywhere that explains the different GHC "groups"

When you say "groups" what do you mean? Do you mean bodies like the core libraries committee, GHC steering committee, etc.?

silky:
> When you say "groups" what do you mean? Do you mean bodies like the core libraries 
> committee, GHC steering committee, etc.?

Yes; it seems to me a lot of time in this thread is being spent on the various domains/interests/roles not being clear? (Could just be my misunderstanding.)

silky:
@hasufell 
> Committees are documented here: https://www.haskell.org/community/

No; that's not a complete list; but thanks, I didn't know about that other issue.

proposals/accepted/051-ghc-base-libraries.rst
michaelpj: That seems like more work for less fine control? If we already have the mechanism in GHC at a module level, that seems sufficient to me.

gbaz: Hackage itself cannot and should not read module-level pragmas, nor can cabal. This proposal is about having some _packages_ be internal and some not. So we need package-level mechanisms for marking things internal, such that both hackage and cabal can take appropriate action (perhaps additional html, like a big red warning box on hackage, and for cabal the ability to warn on using internals as a _direct_ rather than indirect dependency, etc)

hasufell: > and for cabal the ability to warn on using internals as a direct rather than indirect dependency, etc

From what I understood people are also worried about indirect dependencies (coming from things other than the original package), because they may easily lack behind upgrades and bubble up churn.

gbaz: Yes, that would be another thing that would be nice to address. the `category` field does not address that, but seems useful nonetheless, and I would encourage its consideration regardless.

michaelpj: (Sorry, I realised that what I was saying was ambiguous based on the line I was commenting on: I'm in favour of a module-level INTERNAL warning, _added manually to source files_, not magically added to packages by cabal or Hackage)

simonpj:
Thank you Andrew!

Simon

hasufell:
> I'm happy to announce that [the proposal](https://github.com/haskellfoundation/tech-proposals/pull/51/commits/c157f56c4f77ee1c7caaf5c1433296427971ed94) has been [approved unanimously](https://groups.google.com/g/haskell-core-libraries/c/10SyhA9Lm98/m/Zimb2ZZUAAAJ) by CLC.

Sorry to nitpick, but I don't see the vote from @mixphix on the mailing list. Did they reply privately?

mixphix:
Oh, my apologies, I did indeed reply privately by mistake.

hasufell:
I think it is time to merge @Ericson2314 or does the HF want to do anything before that @david-christiansen ?

simonpj:
> I think it is time to merge @Ericson2314 or does the HF want to do anything before that @david-christiansen ?

I agree.

Relatedly, once we have done so, I'd like to start a conversation about naming conventions for modules.  See [this thread](https://github.com/ghc-proposals/ghc-proposals/pull/603#discussion_r1262440387).  Where would be a good place to have that conversation?  As an issue (or a proposal) on this HF github?  On the GHC github?  Where?

david-christiansen:
The TWG agrees that this proposal has been approved by the relevant parties and is useful - we unanimously voted to approve.

david-christiansen:
Thank you all for this process. I think it led to a good outcome!

PR #53, opened 2023-07-25
Module naming conventions for GHC base libraries

simonpj:
This proposal suggests naming conventions for the libraries
* `ghc-internal`
* `ghc-experimental`
* `base`
* `ghc`

[Rendered form here](https://github.com/simonpj/hf-tech-proposals/blob/ghc-module-naming-2023/proposals/0000-ghc-module-naming.rst)


proposals/0000-ghc-module-naming.rst
```diff
@@ -0,0 +1,139 @@
...
+
+
+Proposal 4
+------------
+
+* The public API of package ``ghc`` (GHC as a library) should have modules of form ``GhcAPI.*``.
+
+All of the modules in package ``ghc`` currently start with ``GHC.*`` which correctly signals that they are part of GHC's internals.
+As part of the GHC API redesign (a HF project in its own right, currently stalled) it would be very helpfult
+to modules with stable APIs, and a new prefix, such as ``GhcAPI.*``.
```
Ericson2314, line 87: I am not so sure about this. bifurcating GHC into "public API" vs "internals" is just one way to go about organizing its code (and I personally don't think it is a good one). Regardless of whether it is good or bad, the module naming I think should be independent of such matter of internal organization: modules in GHC should be named the same way whether they are more or less stable.

(If we really had a stable deivions to make, we would have a separate library per https://nikita-volkov.github.io/internal-convention-is-a-mistake/, but we are nowhere near that.)

"API" is also, I think, bad bargain, the "application" and "programming" don't mean anything in particular, and "interface" is quite redundant when all module naming is for interface purposes.

------

`GHC.` I still think is the best name for the compiler proper. GHC stands for Glasgow Haskell *Compiler* --- it's in the name.

The counterparts to `ghc-internals`/`ghc-prim` are called things like `libgcc` (GCC) or `libcompiler-rt` (Clang). The `lib` prefix of C/C++ libraries is doing some work here, `compiler-rt` is rather generic. but I think the `rt` is good. We have "RTS" for just the C part, but it might almost be better to think of the runtime as something encompassing all of `rts` + `ghc-prim` + `ghc-internals`. None of this stuff is part of the compiler proper, but all of it is unstable "support code" propping up the code the user actually wrote. And functionality does in fact move between the various parts of it with some fluidity --- for example @dcoutts is working on moving some IO stuff *back* into C for the threaded runtime (like the unthreaded runtime does today) for more performance [`io_uring`](https://en.wikipedia.org/wiki/Io_uring) support.

So I don't yet have an alternative name I really like, but given the above, maybe something like `GHR` for "Glasgow Haskell Runtime" would make sense.

gbaz: "bifurcating GHC into "public API" vs "internals" is just one way to go about organizing its code" -- I think/hope that the ghc api proposal is _not_ this. Rather, it is to create a _new_ set of modules with _new_ functions that promise to be a stable (and usable) api, and which are _not_ intended for internal use by ghc. That is to say, ghc itself does _not_ dogfood the ghc api (although ghci may choose to do so) and instead the functions provided by the api are designed from the start for external consumption, perhaps glossing over certain tricky but rare bits at first, especially ones highly dependent on less stable datatypes. As such, having a distinct namespace of some sort for the API is very important regardless.

I can imagine such a package perhaps evolving over time to be a distinct package on top of ghc, but that seems orthogonal to this discussion. And, well, now that I think about it, since the API proposal currently is stalled, we probably need not worry about it at all in this discussion -- I just wanted to clarify what the idea is, last I understood it, for the sake of future discussion.

gbaz:
Thanks Simon. All this looks pretty eminently rational to me. I especially like proposal 3 that we shouldn't go around renaming modules that already exist rapidly, but rather should over time have individual discussions as we start to disentangle things and migrate them across packages.

proposals/0000-ghc-module-naming.rst
```diff
@@ -0,0 +1,139 @@
...
+
+* Modules in ``base``, ``ghc-experimental``, ``ghc-prim``, ``ghc-internals`` etc should all have distinct names.
+
+That principle leads immediately to the question: what should those names be?  Hence proposal 2.
+
+Proposal 2
+-----------
+
+* Modules in GHC's internal libraries (``ghc-prim``, ``ghc-internals`` etc) should be of form ``GHC.*``.
+* Modules in ``ghc-experimental`` should be of form ``Experimental.*``.
```
nomeata, line 55: I’d recommend to make the `Experiment` a _suffix_.

Rationale: `Data.Tuple.Experimental` is an companion/extension of `Data.Tuple`; some exports may move from one to the other. Many developers sort their imports alphabetically. Making this a suffix means all `Data.Tuple`-related imports are next to each other.

Ok (omitting explicit import lists and qualifiers):
```
import Control.Applicative
import Control.Arrow
import Experimental.Control.Applicative
import Experimental.Foreign.C
import Data.Tuple
import Foreign.C
```
Better:
```
import Control.Applicative
import Control.Applicative.Experimental
import Control.Arrow
import Data.Tuple
import Foreign.C
import Foreign.C.Experimental
```

Also, maybe people will use the idiom
```
import Data.List qualified as L
import Data.List.Experimental qualified as L
```
which is also nicer if both `qualified as L` are next to each other.

simonpj: Clearly this is a matter of taste.   Personally I find it easier to think of `Experimental.*` as a complete sub-tree of modules, all experimental.  And it's consistent with using a prefix `GHC.*` or `GhcAPI.*` elsewhere.  

But it doesn't matter what I feel provided whatever we do
* Satisfies the maximum number of users
* Is carried through consistently (e.g. all modules in `ghc-experimental` end in `.Experimental`.

I would love to hear from others about prefix-vs-postfix.

proposals/0000-ghc-module-naming.rst
```diff
@@ -0,0 +1,139 @@
...
+
+Proposal 3 only expresses a direction of travel.  We will have to see what the CLC's attitude is,
+and what the Haskell community thinks.  Anything that disturbs the API of base needs to be considered
+rather carefully.
+
+
+Proposal 4
+------------
+
+* The public API of package ``ghc`` (GHC as a library) should have modules of form ``GhcAPI.*``.
```
googleson78, line 83: Is there some particular reason to use a "merged prefix"(`GhcAPI.*`) instead of using the module structure to express the relationship (`GHC.API.*` - this is the `API` of `GHC`)? The latter feels more natural.

nomeata, line 83: This is cosmetics, but I find it strange that suddenly `Ghc` in `GhcAPI` is cased like this.

I would find `GHC.API.` more natural, but I understand that you want to avoid having the public API inside the namespace for the rest of GHC.

There is the `Language.` prefix commonly used by most compiler-like-libraries (e.g. [`haskell-src-exts`](https://hackage.haskell.org/package/haskell-src-exts), [`ghc-parser`](https://hackage.haskell.org/package/ghc-parser), [`ghc-opts`](https://hackage.haskell.org/package/ghc-options)). We could reasonably join this namespace instead of inventing our own, maybe using `Language.Haskell.GHC.*` or, to avoid overly long names, `Language.GHC.*`.

proposals/0000-ghc-module-naming.rst
simonpj: > "bifurcating GHC into "public API" vs "internals" is just one way to go about organizing its code" -- I think/hope that the ghc api proposal is not this. Rather, it is to create a new set of modules with new functions that promise to be a stable (and usable) api, and which are not intended for internal use by ghc. That is to say, ghc itself does not dogfood the ghc api (although ghci may choose to do so) and instead the functions provided by the api are designed from the start for external consumption, perhaps glossing over certain tricky but rare bits at first, especially ones highly dependent on less stable datatypes. As such, having a distinct namespace of some sort for the API is very important regardless.

Yes, @gbaz has it exactly right.   `GhcAPI.*` is a shim around the internal modules `GHC.*`, one that has stronger stability guarantees.  But those internal modules *must* remain available because we can't predict everything that a client of the `ghc` library may want to do.  I hope that if someone finds they can only do something through `GHC.*` they will petition the GHC API working group (still in the womb) to add a suitable function to `GhcAPI.*`.

We must be able to reorganise and refactor GHC's internals without constraint.  At the moment we simply don't know what clients of `ghc` are relying on, so we may well mess up their lives without ever knowing.  If we had `GhcAPI.*` we'd know which modules we needed to take (much) more care with.

The details are not important;  I'm just using this proposal as a way establish, in principle. a namespace for a stable GHC API.

Ericson2314: I like ignoring it for now; I would like it if this proposal didn't mention that at all, and if/when we do that we worry about its module names then.

That brings us back to using `GHC.*` both for the compiler (internals) and for `ghc-internals`/`ghc-prim`. In another thread, this was brought up as a deficiency. This proposal doesn't yet talk about this problem, but shouldn't it?

proposals/0000-ghc-module-naming.rst
Ericson2314: I also think suffix is better. We've had very good luck not having package-distinct prefixes --- I am not sure why practice has gone better than theory! --- so I think it is OK to keep on doing that here.

proposals/0000-ghc-module-naming.rst
```diff
@@ -0,0 +1,145 @@
...
+* We could use ``GHC.*`` for modules in ``ghc-experimental``, and maybe ``GHC.Internals.*`` for module in ``ghc-internals``.  But
+
+  * There are two sorts of GHC-specific-ness to consider:
+
+    * Modules that are part of GHC's implementations
+    * Modules that support a GHC extension, blessed by the GHC Steering Committee
+
+    It is worth distinguishing these: it's confusing if both start with ``GHC.``.
+
+  * It would be a huge upheaval (with impact on users) to rename hundreds of modules in ``ghc-internals``.
```
adamgundry, line 130: Is this really the case? At the moment `ghc-internals` doesn't yet exist, so nobody depends on it! And the whole point is that most users should not depend on it, they should depend on `base` instead!

So why couldn't we establish the convention that `ghc-internals` uses `GHC.Internal.*` as its preferred module prefix, at least for new modules? We could still move over existing modules from `base` without renaming them if that was practically easier, and then rename them later. Renaming modules in `ghc-internals` should be fairly cheap as users aren't supposed to import them anyway, so the only package affected should be `base`.

proposals/0000-ghc-module-naming.rst
Ericson2314: Yes I would prefer it only talk about libraries we've planned on making.

proposals/0000-ghc-module-naming.rst
tek: If we land on `GHC.Internal` and `Foo.Experimental`, I feel like the package should be named `ghc-internal` without the `s`, for consistency

simonpj:
I think that only six people have contributed to this discussion so far.  I bet that many more have an opinion.  Please do say what you think!  Reminder:

This proposal suggests naming conventions for the libraries

* ghc-internals
* ghc-experimental
* base
* ghc

[Rendered form here](https://github.com/simonpj/hf-tech-proposals/blob/ghc-module-naming-2023/proposals/0000-ghc-module-naming.rst)

proposals/0000-ghc-module-naming.rst
tek: I think the reasoning for using a suffix is convincing.

proposals/0000-ghc-module-naming.rst
simonpj: > Is this really the case? At the moment ghc-internals doesn't yet exist, so nobody depends on it! And the whole point is that most users should not depend on it, they should depend on base instead!

I anticipate that *many* of the modules in `base` will move to `ghc-internals`, leaving behind a shim module.  But I suppose it is true that when we move, say `GHC.Base` into `ghc-internals`  we could
* Rename it to `GHC.Internal.Base`, in `ghc-internals`
* Leave behind a shim module in `base` that
  * Is called `GHC.Base`
  * Imports `GHC.Internal.Base` and re-exports it all

Hmm.  That's true.   Moreover, because for a long time (possibly forever) we will have modules like `GHC.Base` in `base` (for back-compat reasons) it would be good if the module in `ghc-internals` had a different name.

simonpj:
> I'm also in favour of the `.Experimental` suffix. I imagine `Data.List.Experimental` could re-export the contents of `Data.List` and add some extra experimental goodies; it's then a very small delta to switch between import Data.List and import `Data.List.Experimental`, whereas import `Experimental.Data.List` would appear somewhere completely different if imports are sorted.

So to be clear, you advocate:

* Use a `.Experimental` *suffix* in `ghc-experimental`
* Use a `GHC.Internal.` *prefix* in `ghc-internals`

Is that right?  It seems superficially inconsistent, and yet each on its own makes sense.

Modules in the `ghc` library (GHC API) are all called `GHC.*` (e.g. `GHC.Tc.Gen.App`) and I don't think we can change that.

What do others think?  It's all about what colour to paint the bikeshed, but we do need to agree it, otherwise we'll get a mess of inconsistent approaches in different modules.

proposals/0000-ghc-module-naming.rst
Ericson2314: I think it's OK if deprecated modules have non-standard names. If anything it is good; it adds an extra signal saying what ought to be used.

adamgundry: > - Leave behind a shim module in base that
>   - Is called `GHC.Base`
>   - Imports `GHC.Internal.Base` and re-exports it all

Yes. Ideally, `GHC.Base` will re-export individual declarations from `GHC.Internal.Base` using an explicit export list (not a `module` export). That way any new definitions added to `GHC.Internal.Base` won't accidentally leak into the `base` API.

So I think it's positively helpful to give the internal modules new names. Another example: we could add module-level `DEPRECATED` pragmas to the `base` shim modules prior to removing them.

adamgundry:
> So to be clear, you advocate:
> 
> * Use a `.Experimental` _suffix_ in `ghc-experimental`
> * Use a `GHC.Internal.` _prefix_ in `ghc-internals`
> 
> Is that right? It seems superficially inconsistent, and yet each on its own makes sense.

Exactly. I'm not particularly worried about making these consistent, since they serve quite different purposes.

proposals/0000-ghc-module-naming.rst
```diff
@@ -0,0 +1,145 @@
...
+Alternatives
+==============
+* We could dispute Proposal 1: one could imagine deliberately naming modules in ``ghc-experimental`` with the
+  same module name as their eventual expected (by someone) home in ``base``.  The goal would be to reduce impact if and when
+  the module moves from ``ghc-experimental`` to ``base``. For example, we might add ``Data.Tuple`` to ``ghc-experimental`` containing the new type constructors ``Tuple2``, ``Tuple3`` etc that are proposed in `GHC Proposal 475 <https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0475-tuple-syntax.rst>`_.   However:
+
+  * In the meantime there are two modules both called ``Data.Tuple``.  This is bad.  Which one does ``import Data.Tuple`` import?  (Look at the Cabal file, perhaps?)  How can I import both?  (Package-qualified imports perhaps.) So it will really only help in the case of a brand-new module, not already in ``base``.
+  * It loses the explicit cue, in the source code, given by ``import Experimental.Data.Tuple``.
+
+* We could use ``GHC.*`` for modules in ``ghc-experimental``, and maybe ``GHC.Internals.*`` for module in ``ghc-internals``.  But
```
michaelpj, line 121: If we adopt the proposal of [having a module warning for the internal modules](https://github.com/haskellfoundation/tech-proposals/blob/main/proposals/accepted/051-ghc-base-libraries.rst#discourging-the-direct-use-of-ghc-internals), then perhaps we don't need a `ghc-internals`-specific module prefix and we could just use `GHC`. Not sure if that's a good idea.

michaelpj:
> Is that right? It seems superficially inconsistent, and yet each on its own makes sense.

I think there is a logic to it:
- `Foo.Experimental` is a close cousin or variant of of `Foo`: you might well want to use both together, and in the long run the latter will hopefully migrate to the former
- `GHC.Internal` and `GHC` may have little to do with each other apart from the fact that the latter uses the former. There might be `GHC.Foo` and `GHC.Internal.Foo`, but more likely there will be completely independent sets of modules.

numlit2:
> Use a `.Experimental` _suffix_ in `ghc-experimental`

`Data.List.Experimental` is not as strong as `Experimental.List`. In a code review with sorted imports scattered `X.Y.Experimental` are easier to miss than a group of `Experimental.*`.

If you want to optimize for experts do `X.Y.Experimental`, if you want to make it more obvious and easier to spot use `Experimental.*`. I think the introduction of unstable dependencies should be as prominent as possible.

tek:
it _would_ be possible to expose both…

nomeata:
> if you want to make it more obvious and easier to spot use Experimental.*

Or `EXPERIMENTAL.` maybe? Too bad we can't include ⚠️ in the module name.

JakobBruenker:
Instead of relying on the module name to warn users, another solution could be to add `{-# WARNING in "x-experimental" Data.List.Experimental "<some message>" #-}`, and have users use `Wno-x-experimental` if they are sure they want to use it. I don't think we currently have module-scoped warnings though.

vanceism7:
I like the idea of X.Y.Experimental, it feels more appropriately categorized and I kind of doubt someone would miss it. A simple search/replace would probably find them anyways. Perhaps we could even add an experimental warning flag that could be triggered on when desired so the compiler could show warnings for anything deemed experimental that was being used in the code.

I'm just bike-shedding here, but it seems like `ghc-experimental` should actually be called "base-experimental". Because the code living in it has a life-goal of moving into base, and users are encouraged (or atleast not forbidden) to use it in their own code.

I think this naming change helps to enforce the policy that users _shouldnt_ be depending on ghc libraries but the base libs instead

proposals/0000-ghc-module-naming.rst
Ericson2314: That is true, but the "broader runtime" (which this is) vs the compiler itself are very different things. I think `GHC.Internals` isn't even different enough (say we wanted to use `GHC.Internals` for extra unstable parts of GHC itself?) but at least it is *something*.

tomjaguarpaw:
> it seems like ghc-experimental should actually be called "base-experimental". Because the code living in it has a life-goal of moving into base

Whose life-goal? If anyone's, it's the goal of GHC HQ. It's not the goal of CLC, who have purview over base. Therefore I don't think the name "base" should be co-opted here.

> users are encouraged (or atleast not forbidden) to use it in their own code

I truly *hope* the name "ghc-experimental" sounds discouraging because I wouldn't like users to depend on this library lightly!

vanceism7:
> Whose life-goal? If anyone's, it's the goal of GHC HQ. It's not the goal of CLC, who have purview over base. Therefore I don't think the name "base" should be co-opted here.
> 
> I truly _hope_ the name "ghc-experimental" sounds discouraging because I wouldn't like users to depend on this library lightly!

`ghc-experimental` definitely sounds like it's something users shouldn't be referencing in their code. Even `base-experimental` sounds pretty scary to me. But the quote from the rendered makes it seem like it's ok to use it


> - Functions and types in here are usually candidates for later transfer into base. But not necessarily: if a collection of functions is not adopted widely enough, it may not be proposed for a move to base.
> - It is user-facing (user are encouraged to depend on it), but its API is less stable than base.

Based on that wording, it sounds like the intent is that the code from experimental eventually matures and takes a home in base, and that it's ok to use it with the warning that it might change on you.

If that's the case, I think `base-experimental` is more appropriate. If it's supposed to be super scary and normal devs shouldn't be depending on it, then I agree with calling it `ghc-experimental`. But `ghc-experimental` sounds like it shouldn't even be user facing

Bodigrim:
> I'm just bike-shedding here, but it seems like `ghc-experimental` should actually be called "base-experimental". Because the code living in it has a life-goal of moving into base, and users are encouraged (or atleast not forbidden) to use it in their own code.

It seems you are reading the purpose of `ghc-experimental` as being a staging ground for inclusion into `base`. I don't think it's correct: the primary goal for `ghc-experimental` is to **experiment** with API required to implement GHC proposals. It may never end up in `base` (after all, it is only aspirational, "ultimate" goal to do so), and it's more likely than not that, building upon results of experiments, the final design approved for inclusion in `base` will be much different from one in `ghc-experimental`.

>  If it's supposed to be super scary and normal devs shouldn't be depending on it, then I agree with calling it `ghc-experimental`. 

It depends whether experiments scare you, right? If you depend on `ghc-experimental`, you agree to participate in an experiment. Sure, this experiment is not supposed to be "super scary", and it is designed to bring positive outcomes in mind, but there is always a chance, and we want consumers to be conscious about it.

The heuristics is simple: `ghc-internal` for GHC developers, `base` for CLC, and `ghc-experimental` for GHC Steering Committee. If there is `ghc` in the package name, it is guided by a community body with GHC in its name. I'm quite strongly opposed to renaming to `base-experimental`.

vanceism7:
> > I'm just bike-shedding here, but it seems like `ghc-experimental` should actually be called "base-experimental". Because the code living in it has a life-goal of moving into base, and users are encouraged (or atleast not forbidden) to use it in their own code.
> 
> It seems you are reading the purpose of `ghc-experimental` as being a staging ground for inclusion into `base`. I don't think it's correct: the primary goal for `ghc-experimental` is to **experiment** with API required to implement GHC proposals. It may never end up in `base` (after all, it is only aspirational, "ultimate" goal to do so), and it's more likely than not that, building upon results of experiments, the final design approved for inclusion in `base` will be much different from one in `ghc-experimental`.
> 
> > If it's supposed to be super scary and normal devs shouldn't be depending on it, then I agree with calling it `ghc-experimental`.
> 
> It depends whether experiments scare you, right? If you depend on `ghc-experimental`, you agree to participate in an experiment. Sure, this experiment is not supposed to be "super scary", and it is designed to bring positive outcomes in mind, but there is always a chance, and we want consumers to be conscious about it.
> 
> The heuristics is simple: `ghc-internal` for GHC developers, `base` for CLC, and `ghc-experimental` for GHC Steering Committee. If there is `ghc` in the package name, it is guided by a community body with GHC in its name. I'm quite strongly opposed to renaming to `base-experimental`.

Got it. I totally agree with the logic. I think some of those finer points should be included in the description of ghc-experimental just to make it a little more clear. Thanks for the explanation!

nomeata:
> The heuristics is simple: ghc-internal for GHC developers, base for CLC, and ghc-experimental for GHC Steering Committee. L

The apparent inevitability of Conway's Law amazes me every time.

simonpj:
I have done a full refresh of this library-module-naming-convention proposal, and pushed the result.

I think we are done.  Any final thoughts?

proposals/0000-ghc-module-naming.rst
```diff
@@ -0,0 +1,188 @@
...
+
+Example lifecycle
+===================
+
+By way of example, consider the ``HasField`` class, which supports overloaded record fields.
+It is currently defined in ``base:GHC.Records``, which is an odd module to have to import.
+Moreover there is
+more than one GHC proposal that suggest changes to its design (e.g. see `GHC Proposal 158 <https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0158-record-set-field.rst>`_); it is not nearly as stable as most of ``base``
+
+If ``ghc-experimental`` had existed we would have put it in ``ghc-experimental:Experimental.Records``.
```
noughtmare, line 171:
```suggestion
If ``ghc-experimental`` had existed we would have put it in ``ghc-experimental:Data.Records.Experimental``.
```

proposals/0000-ghc-module-naming.rst
```diff
@@ -0,0 +1,188 @@
...
+* The public API of package ``ghc`` (GHC as a library) should have modules whose names clearly distinguish them
+  from internal modules.
+
+For example, the public API could have modules of form ``GhcAPI.*``, or ``GHC.API.*``, or ``Language.Haskell.GHC.*`` or something else. The specifics are a matter for the future GHC API working group.
+
+
+
+Timescale
+==========
+The first release of GHC with `ghc-experimental` and `ghc-internal` will be GHC 9.10, which expect to
```
adamgundry, line 160:
```suggestion
The first release of GHC with ``ghc-experimental`` and ``ghc-internal`` will be GHC 9.10, which expect to
```

adamgundry (commented):
Thanks @simonpj, this looks great.

proposals/0000-ghc-module-naming.rst
```diff
@@ -0,0 +1,188 @@
...
+-----------
+
+* Modules in GHC's internal libraries (``ghc-prim``, ``ghc-internal`` etc) should be of form ``GHC.Internal*``.
+* Modules in ``ghc-experimental`` should be of form ``*.Experimental``.
+* Modules in ``base`` should not have either of these forms.
+
+So example we might have
+
+* ``GHC.Internal.Bits`` in ``ghc-internal``,
+* ``Data.Bits.Experimental`` in ``ghc-experimental``a
```
tek, line 74:
```suggestion
* ``Data.Bits.Experimental`` in ``ghc-experimental``
```

proposals/0000-ghc-module-naming.rst
```diff
@@ -0,0 +1,188 @@
...
+more than one GHC proposal that suggest changes to its design (e.g. see `GHC Proposal 158 <https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0158-record-set-field.rst>`_); it is not nearly as stable as most of ``base``
+
+If ``ghc-experimental`` had existed we would have put it in ``ghc-experimental:Data.Records.Experimental``.
+That would have made it clear that the design of overloaded records still evolving.
+Once the design becomes settled and stable, it could move to ``base``, perhaps in a module like ``Data.Records``.
+
+Other similar examples include
+
+* The tuple proposal of `GHC Proposal 475 <https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0475-tuple-syntax.rst>`_
+* The `DataToTag CLC proposal <https://github.com/haskell/core-libraries-committee/issues/104>`_ would have been easier to expose through ``ghc-experimental`` in the first instance.
```
clyring, line 178: I don't think this is a good example; the motivation for that proposal is not convincing unless the _existing_ `dataToTag#` and `getTag` exposed from `ghc-prim` and `base` get improved types.

simonpj: OK. I have replaced this example with the exceptions proposal.

proposals/0000-ghc-module-naming.rst
```diff
@@ -0,0 +1,188 @@
...
+
+Alternatives
+==============
+* We could dispute Proposal 1: one could imagine deliberately naming modules in ``ghc-experimental`` with the
+  same module name as their eventual expected (by someone) home in ``base``.  The goal would be to reduce impact if and when
+  the module moves from ``ghc-experimental`` to ``base``. For example, we might add ``Data.Tuple`` to ``ghc-experimental`` containing the new type constructors ``Tuple2``, ``Tuple3`` etc that are proposed in `GHC Proposal 475 <https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0475-tuple-syntax.rst>`_.   However:
+
+  * In the meantime there are two modules both called ``Data.Tuple``.  This is bad.  Which one does ``import Data.Tuple`` import?  (Look at the Cabal file, perhaps?)  How can I import both?  (Package-qualified imports perhaps.) So it will really only help in the case of a brand-new module, not already in ``base``.
+  * It loses the explicit cue, in the source code, given by ``import Experimental.Data.Tuple``.
```
s-and-witch, line 187:
```suggestion
  * It loses the explicit cue, in the source code, given by ``import Data.Tuple.Experimental``.
```

bgamari:
I, for one, am quite happy with where this proposal ended up. Thanks, @simonpj!

Ericson2314:
I am too.
 
 - [x] Prefix for `ghc-internals`
 - [x] Suffix for `ghc-experimental`
 - [x] GHC API parts are non-normative 

Thanks @simonpj!

proposals/0000-ghc-module-naming.rst
```diff
@@ -0,0 +1,188 @@
...
+-----------
+
+* Modules in ``base``, ``ghc-experimental``, ``ghc-prim``, ``ghc-internal`` etc should all have distinct names.
+
+That principle leads immediately to the question: what should those names be?  Hence proposal 2.
+
+Proposal 2
+-----------
+
+* Modules in GHC's internal libraries (``ghc-prim``, ``ghc-internal`` etc) should be of form ``GHC.Internal*``.
```
TeofilC, line 67:
```suggestion
* Modules in GHC's internal libraries (``ghc-prim``, ``ghc-internal`` etc) should be of form ``GHC.Internal.*``.
```
I think there's a fullstop missing here

gbaz:
merging as approved.

PR #54, opened 2023-07-27
[PROPOSAL] Job posting platforms and moderation

hasufell:
Rendered: https://github.com/hasufell/tech-proposals/blob/job-posting-platforms/proposals/0000-job-posting-platforms.md

----

Community discussions:

* https://discourse.haskell.org/t/hf-proposal-job-posting-platforms-and-moderation/7121
* https://www.reddit.com/r/haskell/comments/15b29gr/hf_proposal_job_posting_platforms_and_moderation/
* https://kbin.social/m/haskell/t/258324/PROPOSAL-Job-posting-platforms-and-moderation-by-hasufell-Pull-Request

### DIFF
--- a/proposals/0000-job-posting-platforms.md
+++ b/proposals/0000-job-posting-platforms.md
@@ -1,0 +1,105 @@
+# Curated list of community job advertisement platforms
+
+## Abstract
+
+Job postings and hiring threads are an important part of any language community. The Haskell Foundation
+should maintain a list of well-established free platforms where companies can advertise positions, while
+also incentivizing a common moderation standard for those platforms by means of inclusion in said list.
+
+## Background
+
+Currently, there are a couple of well-known places where companies can advertise their jobs and positions for free, such as:
+
+* Haskell Discourse: [https://discourse.haskell.org/](https://discourse.haskell.org/)
+* Haskell reddit: [https://www.reddit.com/r/haskell/](https://www.reddit.com/r/haskell/)
+* Haskell weekly: [https://haskellweekly.news/advertising.html](https://haskellweekly.news/advertising.html)
+
+These places have seen a fair amount of growth in job postings, which can also be seen by the following
+data analysis: https://github.com/nh2/haskell-jobs-statistics
+
+The Haskell Foundation should maintain a curated list of such platforms for convenient visibility
+for companies, while also having inclusion requirements/criteria that promote healthy civil conduct.
+
+## Problem Statement
+
+One of the main problems this proposal is trying to solve is the frequent toxic nature of hiring threads,
+which often derail for various reasons:
+
+* someone doesn't like the industry (e.g. blockchain) and starts arguing why no one should ever work in that industry
+* someone doesn't like the company or CEO and goes on a political rant why those are bad people and no one should work in that company
+* someone doesn't like the proposed job conditions and goes on a rant why capitalism is the root of all evil
+* someone feels the need to express their political views and takes a job posting as an opportunity to do so
+
+Just to give a few examples of such threads:
+
+* [https://discourse.haskell.org/t/senior-platform-engineer-tlon-hosting-platform-remote/5132](https://discourse.haskell.org/t/senior-platform-engineer-tlon-hosting-platform-remote/5132)
+* [https://discourse.haskell.org/t/anduril-industries-is-hiring-haskell-developers/7089](https://discourse.haskell.org/t/anduril-industries-is-hiring-haskell-developers/7089)
+* [https://www.reddit.com/r/haskell/comments/uozcvx/haskellers_needed_at_superrare_labs/?utm_source=share&utm_medium=web2x&context=3](https://www.reddit.com/r/haskell/comments/uozcvx/haskellers_needed_at_superrare_labs/?utm_source=share&utm_medium=web2x&context=3)
+* [https://www.reddit.com/r/haskell/comments/ssnw8i/hiring_remote/?utm_source=share&utm_medium=web2x&context=3](https://www.reddit.com/r/haskell/comments/ssnw8i/hiring_remote/?utm_source=share&utm_medium=web2x&context=3)
+
+This behavior of the community and a few engaged individuals has caused damage to the reputation of those
+platforms in industry: some companies actively avoid posting on platforms, where users can leave replies and comments.
+It causes chilling effects even for less controversial places, because for every company there may be people out there
+that dislike them, their industry or their work for whatever reason. And those companies now face potential public shaming
+or reputation damage.
+
+This is purely a moderation issue. So the question is: how can the HF define and incentivize good moderation principles
+on such platforms, without imposing any of those.
+
+## Prior Art and Related Efforts
+
+I have raised my concerns about moderation issues on both reddit and discourse in the past. From the (previous) reddit moderation
+team, I got a reply, but no action was ever taken. The discourse admin team (or whoever is on the 'discourse-admin' alias/ML)
+never replied.
+
+I have also recently contacted Discourse admins in private, which seemed to foster some understanding and awareness of the problematic
+situation.
+
+My attempts to improve the situation by voicing my concerns directly to the appropriate moderation teams have failed.
+As such, I think we can hack our own social behavior through incentivizing a standard set of moderation practices.
+
+## Technical Content
+
+We propose here that the Haskell Foundation maintains a list of platforms, where companies can advertise jobs.
+A platform may be a social media site, a dedicated site for jobs, a forum or a newsletter, etc.
+
+The inclusion criteria is as follows:
+
+* advertising is free of charge
+* the platform has a clear written policy about:
+  - which type of job postings are allowed (e.g. which languages: Haskell/FP... or whether a salary range is a requirement)
+  - which types of job postings are not allowed (e.g. industry XY is disallowed)
+* if the platform allows users to reply or leave comments, then it must:
+  - allow the job poster to request that the thread be locked or replies be disallowed (if the platform permits)
+  - have the following moderation rules:
+    - postings of personal opinions on companies or industries in job postings are considered spam and will be removed
+    - postings that don't show the intention of engaging with the recruiter or asking genuine questions about the job are considered offtopic
+	- generally, job postings are not discussion threads
+
+We note that users who have opinions on politics, companies or industries can voice these in dedicated forums or threads or write personal
+blog posts about it and link to them in a new thread (such as Stephen Diehl's [remarks on blockchain and cryptocurrency](https://www.stephendiehl.com/posts/crypto.html)).
+As such, free speech is maintained.
+
+Any platform that wishes to adhere to these criteria, can be included in the curated list of job posting platforms.
+This proposals gives no details on how this list is implemented or presented.
+
+## Timeline
+
+None.
+
+## Budget
+
+None.
+
+## Stakeholders
+
+* companies looking to advertise jobs
+* social media platforms and other platforms, where jobs can be advertised
+* the moderation teams of those platforms
+* Haskellers looking for jobs
+
+## Success
+
+* publishing of a curated list of platforms, where companies can advertise jobs
+* listed platforms adhere to the moderation standards outlined in this document
+


googleson78:
> postings of personal opinions on companies or industries in job postings are considered spam and will be removed

Is the stipulation here that if I've worked in a company and I think that it's a bad place to work at, I'm not allowed to share my opinion?

hasufell:
> > postings of personal opinions on companies or industries in job postings are considered spam and will be removed
> 
> Is the stipulation here that if I've worked in a company and I think that it's a bad place to work at, I'm not allowed to share my opinion? (and vice versa for encouragement regarding good experiences)

Yes, doing that in the job posting thread would be considered off-topic.

You can do that in a new thread though (in case the platform doesn't consider that off-topic), in your own blog post sphere or somewhere else where it's appropriate (glassdoor?).

maxigit:
My opinion is that job posts are indeed usefu but prone to toxicity.
An easy way to stop toxic comments is to simply lock posts/thread so that comment are not allowed.
The platform just become a billboard and that it. If people need clarification on the job, they can contact the company.
Also it might forces companies to present a better job description.

What I found weird however is when the OP comes from nowhere. Nobody's ever heard of him/she and usually
they are job for "Haskellers" but actually not Haskell job.

hasufell:
> Nobody's ever heard of him/she and usually they are job for "Haskellers" but actually not Haskell job.

Yes, this proposal doesn't impose any restrictions, but demands that platforms write down their policies clearly.

> An easy way to stop toxic comments is to simply lock posts/thread so that comment are not allowed.

Yes, platforms that do that will naturally adhere to the proposal and are eligible to be listed and endorsed by HF.
Also note that it's fine (as the proposal indicates) if platforms simply give the OP the option to have their thread locked.
But there may be reasons they don't want it locked, e.g. if people have questions about the exact nature of the position.

maxigit:
Sorry I mis-understood the goal if this proposal.

PPKFS:
If your interest is to post up an advertisement and then proceed to disable any community involvement, you want a job posting site. Not a community discussion thread.

> So the question is: how can the HF define and incentivize good moderation principles on such platforms, without appearing to impose any of those.

actively calling for something whilst also saying you need to not appear to be doing that something is never a good sign.

> some companies actively avoid posting on platforms, where users can leave replies and comments.  
> And those companies now face potential public shaming or reputation damage.

Maybe if a company doesn't want to face public shaming, it shouldn't be doing things in public that it can be shamed for.

> I have raised my concerns about moderation issues on both reddit and discourse in the past.
> I have also recently contacted Discourse admins in private
> My attempts to improve the situation by voicing my concerns directly to the appropriate moderation teams have failed

Have you considered the possibility that your attempts have failed because this is not what the community wants? This reads to me like "I personally don't want people bashing blockchain or military defence companies and I have tried to convince mods to ban that discussion but it didn't work so now I'm pulling out the big guns".

hasufell:
> Have you considered the possibility that your attempts have failed because this is not what the community wants?

Yes, which is why I'm raising this as a HF proposal.

Also note that companies are part of the "community" (stakeholders).

> Maybe if a company doesn't want to face public shaming, it shouldn't be doing things in public that it can be shamed for.

This is subjective, unless you're talking about illegal activities.

jkachmar:
i feel like using Anduril as an example sort of dilutes your argument, seeing as the fact that they’re pretty much just plain evil is [one of the few things i’ve seen that brings people together](https://www.reddit.com/r/haskell/comments/9765vg/anduril_industries_is_hiring/) in our little community.

---

less tongue-in-cheek, but i’d the HF thinks that policing the discourse in job postings on community fora is a good use of its time i would probably be personally less inclined to advocate for it in either personal & professional contexts.

hasufell:
> less tongue-in-cheek, but i’d the HF thinks that policing the discourse in job postings on community fora is a good use of its time i would probably be personally less inclined to advocate for it in either personal & professional contexts.

You misread the proposal.

The HF will not police anything. They just maintain a list of platforms they endorse for job postings. If a platform doesn't match the criteria, they will be delisted.

The same approach was taken for the "unified installer proposal". Set standards, allow people to adhere to them. If they don't, that's fine too.

hasufell:
> say that someone were to cross-link that job posting to a separate thread on the Haskell subreddit calling attention to the ways in which Badmanduril’s CEO has financially supported political causes that target marginalized communities.

That would be entirely legitimate according to this proposal. The proposal makes this very explicitly clear.

This is **only** about behavior in the job posting thread itself... and *nothing* else. There's no impact on free speech whatsoever.

An alternative approach might be to only list job boards where no discussions/user replies are possible.

Are you disagreeing that there's a huge amount of toxic comments in job posting threads? How many comments on reddit do you see about blockchain being a scam... on job threads? Does that help anyone evaluate their decisions about applying for said company? I don't think so.

ramirez7:
> Management was aware of the toxic nature of reddit and told us to do the hiring elsewhere.

r/haskell (and public internet Haskell forums in general) exists as a place for users to chat, hang out, etc.

If the free advertising wasn't worth it, that's fine. I don't think the HF needs to get involved here to apply pressure to get forum moderators to change their rules to make employers a different class of user.

hasufell:
> But that means HF will be tracking and evaluating how moderation is done on those platforms? That sounds quite demanding?

I agree that might be a shortcoming of this approach.

I also understand that users may see this as controversial and as HF interfering with moderation standards (although not technically true).

There may be other solutions to fix the problem. I consider this as much of an RFC as it is a concrete proposal. I cross posted this on multiple platforms for a reason.

----

I'm very mindful about the reputation of HF and I will not pursue this or push for this if it may harm its reputation.

hasufell:
> Ok, I would still like to see this expanded. Is your main motivation removing these chilling effects?

Yes correct.

> If so, can we have some evidence that these chilling effects exist? You mention that you've experienced it, so that's at least one point. Is there more?

That's a good point. I'm guessing HF could actually source industry opinions about the state of hiring experiences from the company POV. We don't seem to have good data on that.

> Yes, of course I have - I'm an active participant in /r/haskell. Civility is most certainly something I want. Outright removing the option of the responses outlined in this proposal is a very extreme solution to restore civility. For example, what is uncivil about https://discourse.haskell.org/t/senior-platform-engineer-tlon-hosting-platform-remote/5132/4?

I really want to avoid getting into details about specific posts. I'm fine removing those references from the proposal. I'm also not interested in digging through those toxic threads to find a particularly fitting example.

Sufficed to say: posts that basically start with "this company is bad, because..." are causing chilling effects regardless of whether they are true or not. There's a whole spectrum of these posts, some with more, some with a lot less substance and some that are just outright "blockchain is scam" et al.

Under my proposal, you could still write those posts, but you would have to make a new thread.

I don't think you want the moderation team to get engaged in fact checking and reviewing material etc. So the only options are: allow everything or allow none of it in the hiring thread.

PPKFS:
> Are you disagreeing that there's a huge amount of toxic comments in job posting threads?

I would argue that having job posts for producing systems that directly kill other human beings where you cannot raise concerns over this is far more toxic.

hasufell:
> I think there's a sense somehow that things can be "politically neutral" and that therefore it's easy to ban people who bring up politics.
> 

The proposal says nothing about banning people.

> I would suggest that there is always inherent politics and that it's therefore okay to discuss it; and pretending that there isn't really just enforces a certain status quo.
> 

Yes, but there's place and time to discuss things.

> The idea that job postings themselves are neutral and no comments are allowed seems to be in the vein of status-quo-enforcing, and I personally can't support anything of that kind.

So... as GHCup maintainer, I put up certain colors on the homepage (you might have noticed). That's a humanitarian statement against war and breaking of international law. But it can also be considered to be a political one. And there have been two occasions where people demanded for it to be taken down. I did not.

So, I'm not advocating for obscene neutrality. But I've made that statement at a place where I thought it's appropriate (and also asked haskell.org if they disagree).

I don't think people commenting on hiring threads are really generally interested in changing the world. I view it more as an expression of frustration. And it's not helping the community.

silky:
> I don't think people commenting on hiring threads are really generally interested in changing the world. I view it
> more as an expression of frustration. And it's not helping the community.

Personally I find it really helpful.

I also find it reflects a true fact - a lot of Haskell jobs are not for companies that are amazingly socially-beneficial. I think there's a least one way we can start changing that - switch to jobs and companies that are! Clearly, not all or most of us have the options, but for those that do it's one way of engaging in system change that can be very effective.

In other words, yes, maybe it's annoying to see negativity, but let's shift that by changing the job market, not by simply stopping people from commenting :)

david-christiansen:
I'd like to share a thought that I'm not intending as an argument for or against this proposal, but rather as a clarification about the way I understand it:

> I also understand that users may see this as controversial and as HF interfering with moderation standards (although not technically true).

I don't think that "technically true" applies in the realm of community norms and organizations. It seems clear that this proposal is intended to effect a change in moderation rules. It's not the HF saying "You can't do X" (which makes sense - the HF doesn't have that power!), but it is a proposal that the HF should do what it in fact can do in that direction. I think this proposal should be evaluated as one that does in fact attempt to set moderation standards for forums.

Whether that's an argument in favor or against, I'll leave to further discussion and thought.

Baricus:
While I'm not currently looking for a job, I have used multiple of the listed resources to look for postings in the past and keep up to date in the present.  I feel implementing this proposal, especially if doing so shifts all conversation about any posting elsewhere, would encourage me to just skip over any official "job post" threads.  

Most of the value I have gotten out of viewing job postings on the listed sites has come from the community discussion, not the job posting itself.  That includes discussion of the company which posted, rosy or otherwise.  Responses are the upside to job posts when compared to job boards; I get external context outside the company's control.  Without that, I see little benefit to looking at job posts on discussion sites over anywhere else employers can post jobs.  

I feel this proposal wants to carve out an "official Haskell job board" backed by the Haskell Foundation as a subforum/niche within the listed communities.  I'm not involved enough to know whether that's in the foundation's purview but, regardless, I would imagine such a board would be better off as a separate entity, tailored to suit companies rather than general discussion by users.  Posting to a community forum is intended to advertise directly to that community and doing so comes with the risk that the community, or parts of it, will dislike what you offer.

parsonsmatt:
I think it makes sense for the Haskell Foundation to have a list of job boards, so that Haskell programmers and Haskell companies can more easily find each other. So, fully on board here.

I *don't* think that it makes sense to have standards like "my job post won't get criticized on a public forum." I think that's absurd. Reddit is a great place to post jobs. The Haskell Foundation *should* direct people to Reddit - both to make job postings and for programmers to *find* those job postings - and to do otherwise is a disservice. This is true *regardless* of what moderation standards are "allowed" to be on the list.

*As a Haskell programmer that occasionally looks for jobs*, I want to know about *all* of the places I can find job postings.

*As a Haskell employer*, I want to know about *all* of the places I can make job postings.

Now, this story isn't really written for "Generic Haskell Employer." No one makes mean comments or derails the job posts for Mercury or Lumi or Layer 3 Communications. They don't do it for Co-Star, despite it being an astrology app and arguably a scam, nor do they do it for feeld, which is explicitly for nontraditional relationships that many folks find immoral. They do this for blockchain and military companies. So let's just be real here about *who* the stakeholders are in that first bullet point.

Blockchain and military companies want a listing of *safe spaces* so they can post their ads without any discussion (really, they don't want negative commentary reducing the value of their ad - i'm sure they'd love to be derailed by positive discussion and conversation). That's fine! I'm really not opposed to those spaces existing. I'm not opposed to the Haskell Foundation pointing them out, either.

I think everyone gets what they want if the list of job posts is partitioned: "these resources have X moderation profiles and these resources do not." The blockchain/military companies know where they won't get challenged, and everyone else gets to enjoy the listing.

hasufell:
> I think everyone gets what they want if the list of job posts is partitioned: "these resources have X moderation profiles and these resources do not." The blockchain/military companies know where they won't get challenged, and everyone else gets to enjoy the listing.

That sounds reasonable and less controversial.

hasufell:
I think based on the community perception gathered so far, it's clear that these moderation standards are:

* deemed controversial
* are hard to enforce

Additionally, promoting moderation standards through the HF seems potentially "inappropriate".

As such, I'm withdrawing this proposal and will work on the suggestion by @parsonsmatt in a follow up proposal.

TravisWhitaker:
As a person who spent several years recruiting Haskell engineers outside of Badmanduril, and subsequently spent the past six years recruiting Haskell, Nix, and Rust engineers to work at Badmanduril, I have some thoughts on this proposal that may be useful (or at least relevant) to those thinking about what, if any, action the Haskell Foundation should take. These are my own thoughts and opinions as a long time beneficiary of the Haskell community, and do not represent the views of any current or previous employer, real or fictitious.

I am in full-throated agreement with what I believe to be the ultimate intent of this post: elevating the caliber of the public discourse in the Haskell community. Some might think that this idea is vague, and while I'll admit that I'm not sure how to make it more precise, an example might illustrate what I mean: my mail to the Haskell and GHC mailing lists about Badmanduril job opportunities never received responses of the sort that the Problem Statement section of this proposal describes. Indeed, cartoonish statements of the form "Badmanduril is just plain evil and its founder is a crypto-fascist" do not appear on the mailing lists, but do in other forums.

However, I do not think that the Haskell Foundation applying pressure of any kind to the administrators or moderators of any such forum will improve the situation. I believe that some form of the "culture eats strategy for breakfast" meme applies to this situation. If we want the Haskell community to be viewed as professional, I think the attitudes and norms that lead to professional behavior will have to come from the ground up. I do not think that an HF-approved list of spaces where certain kinds of speech are meant to go unchallenged gets us any closer to better etiquette or improved professionalism in the larger community.

Some people may have experienced material negative impacts from unprofessional behavior in the public Haskell community. Perhaps they were unable to convince their team at work to adopt Haskell after others came away with a negative impression, or perhaps they were discouraged from engaging with what could have been an otherwise productive public conversation. When it comes to job postings, this has not been my experience. If anything, the sort of discourse that this proposal is aimed at improves the apparent engagement of my job postings. For each negative public comment, I receive many more job applications, so the response is not as negative as it may appear to onlookers. In fact, there are Badmanduril employees on my own team that became aware of the opportunity because they encountered the public controversy. Public mud slinging and outright calumny in open discourse is part of the cosmic background radiation for us, perhaps those working in the blockchain industry have a similar experience.

An area where I've witnessed negative impacts of unprofessional discourse in the Haskell community comes down to the Haskell Foundation itself, namely funding thereof. My company makes financial contributions to various open source efforts, including the NixOS Foundation. I'd be very interested in funding the Haskell Foundation through my employer, however, public comments from Haskell Foundation leadership have made this very difficult for me to justify to others in my company. Examples include [this post on the initial Haskell Foundation board AMA](https://www.reddit.com/r/haskell/comments/jnwg7i/comment/gb49huo/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button) as well as public comments from leadership [like this](https://twitter.com/yandereidiot/status/1270200542900477955?s=20). I feel this is a good illustration of the material harm that can befall the larger community because of conduct of the sort described in the proposal.

googleson78:
> elevating the caliber of the public discourse in the Haskell community.

It does not seem to me like the main governing body of Haskell endorsing websites which allow non-interactable ad posts in a public forum is elevating the caliber of anything, other than potentially the amount of hires a company doing the (free) advertising will get (and that is doubtful as well, as you yourself have mentioned).

> If we want the Haskell community to be viewed as professional

Personally, I do not want Haskell to be "viewed as professional" (whatever that means by your definition!) if the only path to that is through Badmanduril and cryptoscams. If I could chose to make those disappear at the cost of Haskell I would do so in a heartbeat. To me being professional is very far from the same as "disallowing interactions entirely", and it is definitely not the same as "staying silent when you see something you perceive as wrong from happening".

> An area where I've witnessed negative impacts of unprofessional discourse in the Haskell community comes down to the Haskell Foundation itself, namely funding thereof.

This is entirely based on assuming that HF not taking funding from Badmanduril is "negative impact", which I personally disagree with. Would it not be negative impact for HF to take that funding causing many of the contributors who disagree with this to not want to engage with HF at all? This also seems to assume that the only negative impact that matters is the one cause to HF and the companies in question. 

(By the way, I want to say thanks to @emilypi for publicly taking this stance!)

TravisWhitaker:
> If I could chose to make those disappear at the cost of Haskell I would do so in a heartbeat.

Perhaps this is an example of the “chilling effects” @hasufell mentioned earlier?

> Would it not be negative impact for HF to take that funding causing many of the contributors who disagree with this to not want to engage with HF at all? 

Engaging in the kind of discourse this proposal was aimed at addressing simply caused a different set of people and organizations to not want to engage with the HF at all.

hasufell:
> If I could chose to make those disappear at the cost of Haskell I would do so in a heartbeat. 

This is borderline threatening. I'm not sure if you are aware how this can be interpreted.

----

@david-christiansen this thread has no use for the proposal anymore. Feel free to lock or keep the discussion open at your discretion. But I'm worried what people will continue to post here.

silky:
> Engaging in the kind of discourse this proposal was aimed at addressing simply caused a
> different set of people and organizations to not want to engage with the HF at all.

That's great; that's the intention of holding strong views.


> Statements of the form “I would rather Haskell as a whole not exist than see your 
> organization contributing” seem inconsistent with the HF’s stated goal of driving adoption.

This is a classic misunderstanding of, if I may, the kind that also drives the current climate crises. We don't want _anything_ anymore at "any cost"; i.e. Haskell adoption always was and now should explicitly be morally/socially/environmentally focused.

silky:
FWIW @hasufell re:

> This is borderline threatening. I'm not sure if you are aware how this can be interpreted.

I appreciate your sentiment, but it seems a bit too ironic to be suggesting that in this argument _against the support of any kind of military funding for Haskell_ there would be even the implicit suggestion that it's a threat _towards_ those people! After all, the idea of violence of any kind is the one, I would suggest, the comment is entirely and completely against :)

david-christiansen:
The proposal has been withdrawn - I think that further discussion here is likely to generate more heat than light. @hasufell - thank you for thinking critically about community norms and discussions, and thanks to everyone else for your thoughts.

PR #56, opened 2023-08-28
Split out AST and Parser libraries from GHC

Ericson2314:
### Problem

The community lacks AST and Parser libraries for Haskell that are both self-contained and up-to-date. Experience has shown that there is only way one way to meet each criterion:

- Be used by GHC, so the library cannot fall behind new language development
- Be separate from GHC, so the library is forced to be self-contained

However, no library has so far done both, nor met meet both criteria.

### Solution

The purpose of this proposal is to make that library finally exist. The Haskell Foundation will finance the completion of the existing "Trees that grow" project, decoupling GHC's AST and parser from the rest of the compiler so they can be moved to separate libraries. Those libaries will be "normal" haskell libraries, without any weird dependencies or build process, and published on Hackage. Those libraries will be used by GHC, ensuring they are maintained.

[Rendered](https://github.com/Ericson2314/tech-proposals/blob/62c4acaafe7e994b2d7f349978d83aa26077af15/proposals/accepted/000-ast-parser-libs.rst)

### DIFF
--- a/proposals/accepted/000-ast-parser-libs.rst
+++ b/proposals/accepted/000-ast-parser-libs.rst
@@ -264,25 +264,28 @@ Stakeholders
 
 *Who stands to gain or lose from the implementation of this proposal? Proposals should identify stakeholders so that they can be contacted for input, and a final decision should not occur without having made a good-faith effort to solicit representative feedback from important stakeholder groups.*
 
-- GHC Developers
+GHC Developers
+--------------
 
-  The proposal is asking that we change out code in GHC is organized, so it is crucial that we solicit feedback from the broader `GHC Team <https://gitlab.haskell.org/ghc/ghc-hq/-/tree/main#2-the-ghc-team>`_, and the narrow `GHC HQ group <https://gitlab.haskell.org/ghc/ghc-hq/-/tree/main#3-ghc-hq-group>`_ in particular.
-  It is John's understanding that the GHC developers are broadly supportive of the goal here in the abstract,
-  (after all, SPJ was an author of the Trees That Grow paper),
-  but some of the specific details needed to get this done in a timely manner may be more controversial.
+The proposal is asking that we change out code in GHC is organized, so it is crucial that we solicit feedback from the broader `GHC Team <https://gitlab.haskell.org/ghc/ghc-hq/-/tree/main#2-the-ghc-team>`_, and the narrow `GHC HQ group <https://gitlab.haskell.org/ghc/ghc-hq/-/tree/main#3-ghc-hq-group>`_ in particular.
+It is John's understanding that the GHC developers are broadly supportive of the goal here in the abstract,
+(after all, SPJ was an author of the Trees That Grow paper),
+but some of the specific details needed to get this done in a timely manner may be more controversial.
 
-  In particular, introducing more extension points to ensure rapid progress was very controversial before, and in return for putting up with such a thing as stop-gap, the GHC HQ might want something in return, like an additional phase of work to eliminate the new extension points afterwords.
+In particular, introducing more extension points to ensure rapid progress was very controversial before, and in return for putting up with such a thing as stop-gap, the GHC HQ might want something in return, like an additional phase of work to eliminate the new extension points afterwords.
 
-- Haddock Developers
+Haddock Developers
+------------------
 
-  The Haddock maintainers will likewise be maintaining the result of the Summer of Code project, along with the integration work done as part of this.
-  We should ensure that they are satisfied with the work being done here and it comports with their overall desires for the project.
+The Haddock maintainers will likewise be maintaining the result of the Summer of Code project, along with the integration work done as part of this.
+We should ensure that they are satisfied with the work being done here and it comports with their overall desires for the project.
 
-- HLint Developers
+HLint Developers
+----------------
 
-  The HLint developers have been heavily involved with reusable AST and parser work every step of the way, and should continue to be involved with this too.
-  In addition, we've chosen HLint to be the integration step for the second half just like Haddock was in the first.
-  Thankfully, one of the HLint developers, Shayne Fletcher, is also a co-author of this proposal!
+The HLint developers have been heavily involved with reusable AST and parser work every step of the way, and should continue to be involved with this too.
+In addition, we've chosen HLint to be the integration step for the second half just like Haddock was in the first.
+Thankfully, one of the HLint developers, Shayne Fletcher, is also a co-author of this proposal!
 
 Future Work
 ===========
--- a/proposals/accepted/000-ast-parser-libs.rst
+++ b/proposals/accepted/000-ast-parser-libs.rst
@@ -116,17 +116,24 @@ It can be a good idea to describe alternative approaches here as well, and why t
 
 *How much money is needed to accomplish the goal? How will it be used?*
 
-The project is split into two separate steps: separating the parser, and separating the AST.
+The project is split into two separate steps: separating the AST, and separating the parser.
 Each step has a method, time estimate, and (most importantly) clear success criteria, including use by downstream projects to ensure value is delivered.
 The intent is thus that they are self-contained, and can be individually funded.
 
-Separate the Parser
--------------------
+
+Separate the AST
+----------------
 
 Split library
 ~~~~~~~~~~~~~
 
-**Time Estimate:** ??
+**Time Estimate:** 1--2 Weeks
+
+The first step is just separating data definitions.
+We don't need to worry about code entangling, just data entangling.
+
+The timeline for this is pretty short because there exists an easy last-resort way to decouple anything:
+just add another TTG type family.
 
 Proof of success: Use by Haddock
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -135,11 +142,8 @@ Proof of success: Use by Haddock
 
 https://gitlab.haskell.org/ghc/ghc/-/issues/21592#note_519447 Note how this use-case only needs the AST not parser.
 
-Separate the AST
-----------------
-
-Split library
-~~~~~~~~~~~~~
+Separate the Parser
+-------------------
 
 **Time Estimate:** ??
 
--- a/proposals/accepted/000-ast-parser-libs.rst
+++ b/proposals/accepted/000-ast-parser-libs.rst
@@ -1,0 +1,56 @@
+Haskell Foundation Project Template - place your title here
+===========================================================
+
+*This template is for Haskell Foundation Technical Proposals that are projects to be executed by the Haskell Foundation, without necessarily having any further involvement by the proposer.*
+
+*Please delete the italic text before submitting.*
+
+Abstract
+--------
+
+*This section should provide a summary of the proposal that identifies the key problems to be solved and summarizes the solution.*
+
+Background
+----------
+
+*This section should explain any background (targeting a casual audience) needed to understand the proposal’s motivation (e.g. a high level overview of the technical details and some history).*
+
+Problem Statement
+-----------------
+
+*This section should describe the problem that the proposal intends to solve and how solving the problem will benefit the Haskell community.
+It should also enumerate the requirements against which a solution should be evaluated.*
+
+Prior Art and Related Efforts
+-----------------------------
+
+*This section should describe prior attempts to solve the problem, other relevant prior work, and what others in the community are doing to address the problem.
+It should describe the relationship between the proposed work and the existing efforts.
+If past attempts did not succeed, this section should provide a theory of why not.*
+
+Technical Content
+-----------------
+
+*This section should describe the work that is being proposed to the community for comment, including both technical aspects (choices of system architecture, integration with existing tools and workflows) and community governance (how the developed project will be administered, maintained, and otherwise cared for in the future).
+It should also describe the benefits, drawbacks, and risks that are associated with these decisions.
+It can be a good idea to describe alternative approaches here as well, and why the proposer prefers the current approach.*
+
+Timeline
+--------
+
+*Are there any deadlines that the HF needs to be aware of?*
+
+Budget
+------
+
+*How much money is needed to accomplish the goal? How will it be used?*
+
+Stakeholders
+------------
+
+*Who stands to gain or lose from the implementation of this proposal? Proposals should identify stakeholders so that they can be contacted for input, and a final decision should not occur without having made a good-faith effort to solicit representative feedback from important stakeholder groups.*
+
+Success
+-------
+
+*Under what conditions will the project be considered a success?*
--- a/proposals/accepted/000-ast-parser-libs.rst
+++ b/proposals/accepted/000-ast-parser-libs.rst
@@ -89,7 +89,7 @@ Ultimately, it is those projects we want to help out.
 HLint_, the Haskell linter project developed continuously since 2008 is the most notable one, not just because its longstanding and wide use, but also because many of the developers that worked on the previous two libraries also worked on it --- use by HLint served as proof the libraries were fit-for-purpose.
 
 In June 2019,
-following the release of |ghc-lib-parser| and deprecation of |haskell-src_exts| earlier that year as described above,
+following the release of |ghc-lib-parser| and deprecation of |haskell-src-exts| earlier that year as described above,
 `HLint began the transitition <http://neilmitchell.blogspot.com/2019/06/hlints-path-to-ghc-parser.html>`_ to |ghc-lib-parser|.
 In May 2020, the release of HLint-3.0 which "uses the GHC parser" `was announced <http://neilmitchell.blogspot.com/2020/05/hlint-30.html>`_.
 
--- a/proposals/accepted/000-ast-parser-libs.rst
+++ b/proposals/accepted/000-ast-parser-libs.rst
@@ -170,6 +170,31 @@ We don't need to worry about code entangling, just data entangling.
 
 The timeline for this is pretty short because there exists an easy last-resort way to decouple anything:
 just add another TTG type family.
+This came up with some acrimony in `GHC Issue #21628 <https://gitlab.haskell.org/ghc/ghc/-/issues/21628>`_, discussing whether it was better to try to change GHC's ``FastString`` or abstract over it.
+The purpose of this proposal isn't to relitigate that issue, but because this proposal *is* about resource allocation, something does need to be said on the broader tradeoffs at play
+
+There is no disagreement that as-is, that data type is not suitable for a nice self-contained library. [#faststring-unsuitable]_
+The disagreement is whether TTG should be blocked on reworking ``FastString`` somehow to be better for GHC and non-GHC alike, or whether we should just side-step the issue entirely.
+
+I make no claims about what is better in the long term for GHC, but when reworking ``FastString`` and benchmarking the new algorthms might take `**Days to Weeks**`, we can side-step the issue with a new ``StringP`` type family "extension point" like the existing ``IdP`` one in **minutes**. [#extension-point]_
+
+Out of a basic fiduciary towards the ``Haskell Foundation``, we thus declare that unless "Plan A" works out very quickly, "Plan B" of just introducing another extension point should be used.
+We can also revisit the issue later, *after* we have our factored-out AST library.
+
+.. [#faststring-unsuitable]
+  Everyone agrees it is insuitable in its current state because things like:
+
+  - Global state because of `string interning <https://en.wikipedia.org/wiki/String_interning>`, with a global variable baked into the RTS no less!
+
+  - Memoizing features for other parts of the compiler unrelated to parsing, such as the `"Z-Encoding" <https://gitlab.haskell.org/ghc/ghc/-/blob/261c4acbfdaf5babfc57ab0cef211edb66153fb1/libraries/ghc-boot/GHC/Utils/Encoding.hs#L43>` GHC happens to use for object file symbol `name mangling <https://en.wikipedia.org/wiki/Name_mangling>`.
+
+  Everyone *also* agrees that it is worth revising whether these algorithmic decision still make sense given modern hardware, see `GHC Issue #17259 <https://gitlab.haskell.org/ghc/ghc/-/issues/17259>`_.
+
+.. [#extension-point]
+  "Extension point" is Trees That Grow parlance for such a type family.
+  The idea is that the AST library no longer refers to a data type like ``FastString`` directory, but instead refers to an abstract ``StringP p``.
+  Then, GHC can define `StringP (GhcPass _) = FastString`` to use it client side, across all compilation passes.
+  All term-level code continues to works exactly the same as before without modification.
 
 Proof of success: Use by Haddock
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/proposals/accepted/000-ast-parser-libs.rst
+++ b/proposals/accepted/000-ast-parser-libs.rst
@@ -78,7 +78,8 @@ The lesson from |haskell-src-exts| are clear:
 Whereas |ghc-lib-parser| succeeds in keeping up with GHC because it *is* GHC, it fails in being self-contained because modularity cannot be `"fixed in post" [production] <https://tvtropes.org/pmwiki/pmwiki.php/Main/FixItInPost>`_.
 Code that is intended to be separate from any one consumer must be developed with those boundaries enforced during development.
 
-.. [#ghc-inception] The extraction process was enabled by insights gained from the `"GHCinception" <https://mgsloan.com/posts/ghcinception/>`_ or "GHC in GHCi" initative.
+.. [#ghc-inception]
+  The extraction process was enabled by insights gained from the `"GHCinception" <https://mgsloan.com/posts/ghcinception/>`_ or "GHC in GHCi" initative.
 
 Downstream projects
 -------------------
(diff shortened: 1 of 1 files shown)


alt-romes:
Yes yes yes!

To the section describing the GHC’s developer support for this plan,
I’d be good to add a paragraph linking to the design-focused issue where we’ve reached an initial consensus about exactly the splitting of the AST and parser into e.g. `language-haskell` and `language-haskell-parser`:

https://gitlab.haskell.org/ghc/ghc/-/issues/21592

dpwiz:
+stakeholders: library authors who want to embed (a subset of) Haskell into their DSLs. 

Like templates, interpolation quasiquoters, etc. `haskell-src-exts` is still a frequent dependency of those.

dpwiz:
+stakeholders: library authors who want to generate valid Haskell expressions, declarations and modules without messing with text-level builders.

Ericson2314:
OK @alt-romes, added some bit on that. If you would like some more, feel free to propose an edit! ;)

proposals/accepted/000-ast-parser-libs.rst
```diff
@@ -0,0 +1,344 @@
...
+- Be separate from GHC, so the library is forced to be self-contained
+
+However, no library has so far done both, nor met meet both criteria.
+
+Solution
+--------
+
+The purpose of this proposal is to make that library finally exist.
+The Haskell Foundation will finance the completion of the existing "Trees that grow" project, decoupling GHC's AST and parser from the rest of the compiler so they can be moved to separate libraries.
+Those libaries will be "normal" haskell libraries, without any weird dependencies or build process, and published on Hackage.
```
david-christiansen, line 34:
```suggestion
Those libraries will be "normal" Haskell libraries, without any unusual dependencies or build processes, and published on Hackage.
```

proposals/accepted/000-ast-parser-libs.rst
```diff
@@ -0,0 +1,344 @@
...
+The idea that they are "promoted" via ``DataKinds``, and then type families used in the AST will have instances for these promoted values.
+This allows those consumers to "adjust" the AST for their purpose.
+
+The Trees That Grow project is now 6 years old, and has met great success in avoiding partiality in the compiler, "making illegal states unrepresentable" as many Haskellers would put it.
+But progress on `reducing AST & parser dependencies <https://gitlab.haskell.org/ghc/ghc/-/issues/19932>`_ has been less easily forthcoming.
+We have separated out the modules defining the AST under ``Language.Haskell.Syntax.*`` we wish to split out, and we have tests to track progress reducing their deps, and the parser's deps.
+But progress is unsteady and unpredictable.
+
+The basic problem is that the benefits don't actually kick in until the deps are *all* gone, and the code is actually separated out.
+Partial progress isn't really directly useful to anyone, and these counters just scoreboard by which we hope to get closer to the end goal.
```
david-christiansen, line 145: What does "these counters just scoreboard by which" mean?

proposals/accepted/000-ast-parser-libs.rst
```diff
@@ -0,0 +1,344 @@
...
+The Haskell Foundation's support in getting this "over the finish line", at which point the community *will* benefit, and benefit greatly, is thus a crucial way we can surmount the coordination failure the lack of incremental payoff causes.
+
+.. [#intra]
+  It might sound like the goal is only different usages within GHC, but remember that ``template-haskell`` is a separate library used by users of Haskell not just developers of Haskell.
+  A goal of at least some usage outside GHC was always there.
+
+Reinstallable GHC Lib
+---------------------
+
+One of the problems ``ghc-lib-parser`` aims to solve is that ``ghc`` the library is current cumbersome to install as a regular haskell library (as opposed to by switching toolchains entirely).
```
david-christiansen, line 157:
```suggestion
One of the problems ``ghc-lib-parser`` aims to solve is that ``ghc`` the library is current cumbersome to install as a regular Haskell library (as opposed to by switching toolchains entirely).
```

proposals/accepted/000-ast-parser-libs.rst
```diff
@@ -0,0 +1,344 @@
...
+There is currently work in flight to solve that.
+One that is done, projects like HLint_ *could* just depend on ``ghc`` directly, and still be easily buildable (with Cabal / with Stack / from Hackage) as today.
+
+Just doing this isn't a good solution though, because ``ghc`` exposes a much a wider surface area than what these projects actually want.
+For stability's sake, it is better that those libraries dependent on narrower parsing / AST libraries that only provide what they need.
+And longer term, we hope the "tug of war" of between GHC and these projects as consumers of those libraries, versus just the others having to deal with whatever GHC does with just itself in mind, will result in a higher-quality, more flexible, and overall friendlier library.
+
+In `this comment <https://gitlab.haskell.org/ghc/ghc/-/issues/14409#note_506489>`_, it is suggested that factoring out the AST and parser can be a good first step making a more modular in GHC in general.
+This proposal wish to *stay neutral* on the merits of such a future direction, but it would be remiss not to at least highlight it as one possible outcome.
```
david-christiansen, line 166:
```suggestion
This proposal wishes to *stay neutral* on the merits of such a future direction, but it would be remiss not to at least highlight it as one possible outcome.
```

proposals/accepted/000-ast-parser-libs.rst
```diff
@@ -0,0 +1,344 @@
...
+This came up with some acrimony in `GHC Issue #21628 <https://gitlab.haskell.org/ghc/ghc/-/issues/21628>`_, discussing whether it was better to try to change GHC's ``FastString`` or abstract over it.
+The purpose of this proposal isn't to relitigate that issue, but because this proposal *is* about resource allocation, something does need to be said on the broader tradeoffs at play
+
+There is no disagreement that as-is, that data type is not suitable for a nice self-contained library. [#faststring-unsuitable]_
+The disagreement is whether TTG should be blocked on reworking ``FastString`` somehow to be better for GHC and non-GHC alike, or whether we should just side-step the issue entirely.
+
+We make no claims about what is better in the long term for GHC, but when reworking ``FastString`` and benchmarking the new algorthms might take **Days to Weeks**, we can side-step the issue with a new ``StringP`` type family "extension point" like the existing ``IdP`` one in **minutes**. [#extension-point]_
+
+Out of a basic fiduciary duty, we thus declare that unless "Plan A" works out almost as quickly, "Plan B" of just introducing another extension point should be used.
```
david-christiansen, line 207: I don't think this is a fiduciary duty - those are duties incurred by virtue of being someone else's financial representative, where you must represent their interests. I'd say "To keep costs under control" or similar

david-christiansen: Why does this need to be an extension point with a type family, rather than parameterising the AST over some arbitrary type of strings?

proposals/accepted/000-ast-parser-libs.rst
```diff
@@ -0,0 +1,344 @@
...
+Note that HLint does use a few other things behind the AST and Parser that currently make it into |ghc-lib-parser|, but which we might want in our new libraries.
+
+#. It uses GHC's multi-purpose ``Outputable`` instead of some more dedicated exact-printing machinary
+
+#. It uses ``parseDynamicFilePragma``, and thus GHC's infamous ``DynFlags`` to support pragmas like ``{-# LANGUAGE ... #-}`` and ``{-# OPTIONS_GHC ... #-}``.
+
+For the first case, we might consider factoring ``Outputable`` into a separate library too.
+Or we can prioritize a more dedicated exact-print solution to use instead of ``Outputable`` (see the future work section).
+
+For the second case, we might have to do something temporary like e.g. continuing to use an auto-extracted library liek |ghc-lib-parser|, but depending on our newly factored-output libraries, to get this functionality for HLint_.
```
david-christiansen, line 287:
```suggestion
For the second case, we might have to do something temporary like e.g. continuing to use an auto-extracted library like |ghc-lib-parser|, but depending on our newly factored-output libraries, to get this functionality for HLint_.
```

goldfirere:
I would love to see some text explaining why this is desired. I do see a bunch of people above excited for this, so that's encouraging. But why, concretely?

Specifically, what is problematic about depending on `ghc` as a whole in order to get the parser? Is there worry about upgrades with every GHC release? If so, I don't see this proposal as helping -- I would expect the parser library to change with every release, too. Or is the goal simply for the abstract benefits of being self-contained? That's nice, I agree, but I don't think it would make for the best use of the HF's resources. I do see the section in the proposal about `ghc-lib-parser` and how that falls short. The section mentions the challenge of having a nice interface to the library. Is that all that's needed? Accomplishing that goal seems much easier than all the work described here.

Maybe put most directly: how would this proposal make working and developing in Haskell better?

Separately, the proposal is quiet on the costs of modularity. That is, separating out the parser and its datatypes from the rest of GHC imposes a small but constant tax on GHC development. This is already evident in the separation of the `Language.Syntax.*` modules, where I now need to look in two places to understand what is being stored in our AST (in `Language.Syntax` for the base type and in `GHC` for the extension points). This is not so terrible, but it is a real cost and annoyance. With the current proposal, a GHC author would have to modify an external library in order to change the parser or parse tree. Then rebuilding GHC is likely to be considerably slower than it is today, where one can just rebuild a module or two (if datatypes don't change) and then link and then test. This is another form of tax. I'm not stating here I would take the status quo over the future described in the proposal, but I want to make sure we're considering the ongoing costs of this modularity, not just the execution time to get us from here to there.

simonpj:
> With the current proposal, a GHC author would have to modify an external library in order to change the parser or parse tree. Then rebuilding GHC is likely to be considerably slower than it is today, where one can just rebuild a module or two (if datatypes don't change) and then link and then test

Is that really true?  I was assuming that the new packages would be part of the GHC repo, and would take no longer to build than today.

In principle, more modularity is a good thing.  But
* The proposal would be stronger if it was clearer about who, precisely, will benefit.  Is there a constituency that really wants this?  Can we name them and get their views?
* And why do they want it?  They can do it today, by depending on `ghc`. So is the primary (sole?) benefit that they'll be able to depend on a smaller package?
* I suspect that for many clients, having a parsed tree is not enough: they want a renamed tree or a typechecked tree.  For them, this proposal won't deliver any benefits.
* Eliminating all the dependencies of Language.Haskell.Syntax has proved troublesom.  FastString is a major example; I'm not sure if there are others.   (E.g. there are some `NoGhcTc` calls in the allegedly-ghc-independent tree.)  Working out what to do about these is a blank cheque; I do not know how long it would take.  It's not just a routine matter.
* Does the parser package you envisage include all source-location info? Haddock?  It has to, if GHC is to use it.  But that means populating those extension fields in a parser-specific way.  Maybe that's ok, but it'd be good to lay it out.

Like @goldfiere I'm willing to be convinced, but currently sceptical that this is the top priority in our resource-limited world.

silky:
> Specifically, what is problematic about depending on ghc as a whole in order to get the parser? 

image

?

tomjaguarpaw:
I'm confused. What does this image indicate?

silky:
An attempt to suggest that perhaps part of the problem of depending on `ghc` itself is that it's a massive library.

tomjaguarpaw:
But it's built in, isn't it? You can't *not* have it on your Haskell system.

googleson78:
> Is that all that's needed?

Additionally, the library must be "reinstallable" (if I'm interpreting the term correctly), i.e. it should not be tied to the `ghc` you have installed, much like `ghc-lib-parser` itself is.

> With the current proposal, a GHC author would have to modify an external library in order to change the parser or parse tree. Then rebuilding GHC is likely to be considerably slower than it is today, where one can just rebuild a module or two (if datatypes don't change) and then link and then test.

I'm not too intimately familiar with Hadrian, and I'm not saying this is a trivial or easy change to it, but it should definitely be possible to express dependencies at the cross-package module level, i.e. if only `Foo` from package `A` depends on `Bar` from package `B`, and if `Bar` changes, you don't have to rebuild all the modules from `A`, and instead you only have to rebuild `Foo`.
I'm making the "it should definitely be possible" claim because, afair `rules_haskell` is already capable of doing this.

shayne-fletcher:
> > Is that all that's needed?
> 
> Additionally, the library must be "reinstallable" (if I'm interpreting the term correctly), i.e. it should not be tied to the `ghc` you have installed, much like `ghc-lib-parser` itself is.

this is it, the key motivation for ghc-lib-parser.

from https://github.com/digital-asset/ghc-lib:
"The [GHC API](https://hackage.haskell.org/package/ghc) allows you to use the [GHC compiler](https://www.haskell.org/ghc/) as a library, so you can parse, analyze and compile Haskell code. The GHC API comes preinstalled with GHC, and is tied to that GHC version - if you are using GHC 8.6.3, you get version 8.6.3 of the API, and can't change it. The ghc-lib project solves that problem, letting you mix and match versions of the GHC compiler and GHC API."

e.g. the most recent ghc-lib-parser consists of ghc-9.6.1 source and so HLint can be built with compiler versions in the range [9.2.2, 9.6.1].

proposals/accepted/000-ast-parser-libs.rst
michaelpj: I think part of the point of TTG is that you just have _one_ parameter to the AST type, and all the other bits hang off that with type families. Otherwise you would have a truly ungodly number of type parameters.

michaelpj:
Here's a comment I wrote on the GHC issue tracker that covers this a bit: https://gitlab.haskell.org/ghc/ghc/-/issues/14409#note_506489

Excerpting:

> In particular, since the GHC parser is backwards compatible (mostly), it's generally possible to get away with just using the latest ghc-lib-parser, thus avoiding lots of CPP. To be clear, this means that you can build a library with GHC 9.2 and use the GHC 9.6 parser and use it to parse Haskell source of all versions up to 9.6. Which is pretty great!

Let me elaborate on that a bit. Suppose you want to use `ghc` to parse some Haskell source. You also want to support multiple versions of GHC. That forces you to support multiple versions of `ghc`, which means: lots of CPP. This is not normally a problem when depending on packages - you just say "this only works with the latest major version of `foo`" and you're done. But with `ghc` you are _forced_ to use the old versions of `ghc` if you want to support older versions of GHC. So: CPP everywhere.

`ghc-lib-parser` (and a hypothetical reinstallable GHC parser package) take us back to the normal world. I can just say that my package depends on `ghc-parser >= 9.6` and not have to worry about what version of GHC my user is using.

Of course, this is only true because the GHC parser tends to be backwards-compatible, so you can get away with using the latest version. If you want to rename, typecheck, etc. a program, then you really do want to mimic the GHC that the user has so you really do want an exactly matching `ghc`. This is the situation that HLS is in. So this work won't directly benefit HLS (except insofar as it makes it less painful to write tools that HLS might then use). 

However, I will say that I don't know of any other tool apart from HLS that relies on anything past parsing. I think there's a simple reason for that: parsing can be done (mostly) standalone, whereas to go much further you actually need a proper GHC session with a package database and so on. Getting that all set up is pretty non-trivial and is the whole raison d'etre of `hie-bios`. So I presume that everyone else just takes a look and doesn't even try.

----

I think it's worth emphasizing that basically all the "big" tools for working with Haskell source are now using `ghc-lib-parser`, often migrating to it from `ghc`. I think that's a pretty clear signal of desirability that:
- People are willing to maintain the (I say this with great love) awful, fragile hack that is `ghc-lib-parser`
- People are keen to migrate to `ghc-lib-parser`

michaelpj:
Another thing: it seems to me that the current situation with `ghc-lib-parser` is somewhat dangerous. There are two problems:
1. It requires stitching together various bits of GHC sources. This is fragile and there are 0 guarantees from GHC that it won't break. I have seen several tickets on the GHC issue tracker about changes that broke `ghc-lib-parser`, which the GHC devs are understandably oblivious towards.
2. It's reliant on Shayne doing the (I imagine) fairly unpleasant work of maintaining it. (sorry Shayne I'm putting words in your mouth here, please do correct me).

And since lots of tools now rely on `ghc-lib-parser`, if it stops being updated for some reason... there just won't be any Haskell tools for new GHC versions for a while.

In contrast, a GHC parser package maintained by the GHC team is going to be robustly tested, maintained by the right people, and probably just much less work overall.

proposals/accepted/000-ast-parser-libs.rst
Ericson2314: in this paragraph I am trying to contrast the smooth function of module count hopefully going down with the step function of "benefits only kick in when count hits 0". It's good to track getting closer to 0, but important to remember partial progress != partial benefit.

Perhaps this whole paragraph is confusing and should be reworded?

Ericson2314:
> I was assuming that the new packages would be part of the GHC repo, and would take no longer to build than today.

This is the intent of the proposal. There should be no major downsides; no new submodules to bump, and still loading everything in a single `ghci`. I've added some text to elaborate on this and other things that must *not* happen.

> E.g. there are some `NoGhcTc` calls in the allegedly-ghc-independent tree.

> Eliminating all the dependencies of `Language.Haskell.Syntax` has proved troublesome. `FastString` is a major example; I'm not sure if there are others. (E.g. there are some `NoGhcTc` calls in the allegedly-ghc-independent tree.) Working out what to do about these is a blank cheque; I do not know how long it would take. It's not just a routine matter.

`NoGhcTc` is indeed something that shouldn't be here, but *not* something we need to fix, because it doesn't impose `import`s on the rest of GHC. If we narrowly focus on what is causing violating imports, and not what *ought* to go where in an ideal world, I think this becomes much more tractable.

I indeed don't' want the HF to write a blank check.

- Does the parser package you envisage include all source-location info? Haddock? It has to, if GHC is to use it. But that means populating those extension fields in a parser-specific way. Maybe that's ok, but it'd be good to lay it out.

I think the answer is "yes" --- we definitely want a parser we can resuse. That said, I do not know the Parser parts as well as I know the AST parts.

------

> Specifically, what is problematic about depending on `ghc` as a whole in order to get the parser?

I don't think I am the right person to answer this; I have song strong biases towards modularity but this gets myself and the GHC devs into "we agree to disagree" territory. I hope everyone else in this thread will be able to find a resolution here without me.

I will say this: there are a number of build system and other "dev ops" issues with `ghc` today. The "reinstallable GHC" project is supposed to solve them all, eventually. The `ghc-lib` library also solved them right away in an (everyone agrees) less than ideal manner.

The `ghc-lib` developers could have stopped there, **but then they also went and made `ghc-lib-parser`**. And I think `ghc-lib-parser` is much more widely used than `ghc-lib`. I think understanding why `ghc-lib-parser` was made in addition, and why people prefer to use it, will be a key part of asking these "for whom" and "why" questions.

goldfirere:
Thanks for the responses above. I now better understand why we want this: we want a Haskell program built with (say) GHC 9.2 to be able to parse Haskell files that use syntax from (say) GHC 9.6. Good. I'm on board with that bit of motivation, otherwise known as making the library *reinstallable*.

I also understand that one problem with the existing `ghc-lib-parser` is that there's no guarantee it won't break. What about a proposal that addresses this, by (say) adding some CI to GHC that tries to build `ghc-lib-parser`. Would that be sufficient? I don't know whether the GHC devs would like that, but would that satisfy the goals of this proposal? I'm just trying to understand the minimal action we could take to satisfy the need here.

Regarding downsides: I remain unconvinced about build times. My read of this proposal is that the new parser would live in the same repo as GHC (not in a submodule) but would be its own Haskell package, with its own cabal file. My best understanding of our build tools suggests that changing a file in one cabalized package requires a rebuild of the package, which would then require rebuilds of other packages that depend on the first one. Maybe these dependencies are at the module level, not the package level? But I'd need someone to assert that pretty clearly. (Note: this is not at all about repo organization or anything involving git.)

Ericson2314:
> Thanks for the responses above. I now better understand why we want this

Yay!

> What about a proposal that addresses this, by (say) adding some CI to GHC that tries to build ghc-lib-parser. Would that be sufficient? I don't know whether the GHC devs would like that, but would that satisfy the goals of this proposal? I'm just trying to understand the minimal action we could take to satisfy the need here.

I am not sure that is actually less work. But happy to discuss the next bit first before returning to this.

>Regarding downsides: I remain unconvinced about build times. My read of this proposal is that the new parser would live in the same repo as GHC (not in a submodule) but would be its own Haskell package, with its own cabal file. My best understanding of our build tools suggests that changing a file in one cabalized package requires a rebuild of the package, which would then require rebuilds of other packages that depend on the first one. Maybe these dependencies are at the module level, not the package level? But I'd need someone to assert that pretty clearly. (Note: this is not at all about repo organization or anything involving git.)

With "vanilla" Cabal and cabal-install, what you say is true. These things invoke `ghc --make` to build the entire library in one go, and when an upstream library is changed, in general the downstream library needs to be rebuilt entirely.

Hadrian (and GHC's old make build system) are different however, they invoke GHC in one-shot mode, so there is one process per Haskell module. They do not force all downstream packages to be rebuilt, at least last I heard.

You can try this out by modifying a file of `ghc` with `--freeze1`, and then seeing that not all `ghc-bin` modules need to be rebuilt. The same level of not rebuilding `ghc-bin` when `ghc` changes we'd want for not rebuilding `ghc` (and `ghc-bin`) when the new parser and ast libraries changes.

The `ghc` vs `ghc-bin` situation hopefully means we are *not* breaking any new ground here. :)

michaelpj:
> I'm on board with that bit of motivation, otherwise known as making the library reinstallable.

Another point here - we've talked a bit about reinstallable `ghc`. That's definitely also nice to have, but it's worth noting that it's _much_ more complicated, because we want it to be a _functional_ `ghc`, and that [can be tricky](https://gitlab.haskell.org/ghc/ghc/-/issues/19896#note_406381). On the other hand, a hypothetical `ghc-parser` package would likely be a very boring standard Haskell package and not lead to any of the above head-scratching.

> Regarding downsides: I remain unconvinced about build times.

If we keep pushing on the modularity we can get to a better place with plain cabal build times also. Conceptually, most of GHC shouldn't depend on the parser, only the driver that puts it all together. So if you have the driver as a separate component that depends on the parser, typechecker etc, then changing the parser should mean rebuilding only the driver component.

> What about a proposal that addresses this, by (say) adding some CI to GHC that tries to build ghc-lib-parser. Would that be sufficient?

I think it would help, but I do think a significant part of the problem is that it's not being maintained by the GHC devs. If the GHC devs want to adopt and maintain `ghc-lib-parser`... I guess that would be okay, but it really feels like taking the painful route.

goldfirere:
Thanks again. I'm becoming more convinced. For me the story runs like this:

- There is strong community demand for `ghc-lib-parser`.
- `ghc-lib-parser` is painful to maintain and fragile.
- This proposal describes how we can get to an alternative parser library that is maintained for free (as always, "for free" means "paid for by someone else", in this case the GHC team, in the course of their normal work)
- The proposal also contains a roadmap for getting there (which I have not thought deeply about)

Is that an accurate telling of the story? This story need not be what inspires you (for any definition of "you"), but if it's an accurate story, then it might be what inspires me. (Specifically, some people have a great desire for more modularity. That motivation speaks to them. But, on its own, it does not speak to me, other than as an aesthetic "nice to have".)

proposals/accepted/000-ast-parser-libs.rst
david-christiansen: I think what threw me was the lack of the words "are" and "a" surrounding "just" :-)

```suggestion
Partial progress isn't really directly useful to anyone, and these counters are just a scoreboard by which we hope to get closer to the end goal.
```

proposals/accepted/000-ast-parser-libs.rst
david-christiansen: Sounds good to me

Ericson2314:
@goldfirere That's exactly right!

For *me personally* modularity is very exciting, but I indeed there is not yet enough evidence that other people care enough for that to motivate an HF proposal on alone --- I knew I couldn't just write a GHC modularity in its entirely HF proposal and expect it to go through.

I turned to `ghc-lib-parser` because there was clear community interest in that library. Perhaps people prefer `ghc-lib-parser` to `ghc-lib` because it is more modular, perhaps people prefer it simply because fewer modules => faster builds. I don't know the answer, but per your story that answer doesn't really matter. It's

1. A popular library
2. Used for tasks the community also cares about (more tools!)
3. Currently very tedious for @shayne-fletcher and others to maintain

And the official goal the proposal is (3), while also making the library hopefully also better than it was before, helping out (2). Any modularity benefits are just icing on the cake we need not worry about if folks don't really care for icing :).

Ericson2314:
(Once could state that I am cynically laundering modularity work under something more popular. I prefer to state that I am trying to demonstrate modularity being useful to help solve existing problems, problems that (at least as I understand it) can't very well be solved any other way. Modularity is the *means* but not the *end* of the proposal.)

tomjaguarpaw:
> Once could state that I am cynically laundering modularity work under something more popular

I think you're scoping out work that has benefits across a number of different axes!

Ericson2314:
I certainly hope so! But some of the more abstract ones are hard to argue in an ironclad objective way. I'm fine with leaving them out of it and hoping the main more concrete enough ones are enough to carry the day.

If this goes as well as I hope, it will provide useful a shared experience to ground discussing those other things. But until then those conversations are rather unmoored and I'm OK waiting.

brandonchinn178:
As a maintainer of Fourmolu, I don't have much to contribute beyond what @mrkkrp said. As Fourmolu is a fork of Ormolu, we mostly inherit ghc-lib-parser from Ormolu and don't interact with that dependency a ton.

But as a maintainer of tasty-autocollect, I can't use ghc-lib-parser because it's a compiler plugin, so I have to use ghc directly. So I do use CPP to provide a common interface that's implemented by each GHC version. Even if ghc-lib/ghc-lib-parser could be reused, I'd probably avoid it anyway because 1) long build time and 2) my usecase is small enough that manually writing shims isn't too much work. But if there were a library broken out that can be used for plugins + provides a compatible interface for multiple versions of GHC + is prebuilt/doesnt have a huge build time, it would be useful for tasty-autocollect

mihaimaruseac:
As a maintainer of `hlint`, I also think having the parser as a separate library is worthwhile. Note though that `hlint` migrated from `haskell-src-extras` recently because this library was lagging behind GHC development, so I wouldn't want the same to happen with a separated `ghc-lib-parser` (though I'm optimistic that the fact that the library would still be part of GHC tree means it won't lag behind).

PR #62, opened 2023-12-19
RFC: Reinvigorating the Haskell Symposium

jgbm:
This is an RFC regarding planned changes to the Haskell Symposium: inviting new types of contribution, and submitting accepted papers to JFP rather than having a proceedings.

[Rendered.](https://github.com/haskellfoundation/tech-proposals/blob/4a422513c3523a7eb36d3a7be47c42a427b64f91/proposals/0000-haskell-symposium-changes.md)

### DIFF
--- a/proposals/0000-haskell-symposium-changes.md
+++ b/proposals/0000-haskell-symposium-changes.md
@@ -24,28 +24,18 @@ Talk proposals will not be distributed to attendees, but authors of talk proposa
 
 ## Proposal 2: JFP Submission of Accepted Papers
 
-The steering committee proposes to cease publishing the proceedings of the Haskell Symposium as an archival ACM publication.   Instead, **authors of accepted research papers and functional pearls will have the option of submitting their papers to an expedited review process for inclusion in the Journal of Functional Programming**. This provides a better-regarded publication venue for work presented at the Haskell Symposium, encouraging submission of high quality work to the Symposium.
+The steering committee proposes that authors of accepted papers may pursue publication in the Journal of Functional Programming, with an expedited process, instead of in the proceedings of the Haskell Symposium.  This provides a better-ranked publication venue for work presented at the Haskell Symposium, encouraging presentation of high quality work at the Symposium.
 
 ### For authors
 
-Under this proposal, authors will still submit papers to the Haskell Symposium, and receive reviews from the PC, on the usual schedule.    
+Under this proposal, authors will still submit papers to the Haskell Symposium, and receive reviews from the PC, on the usual schedule.  At the time of submission, authors may indicate that they would prefer their papers be considered for JFP, rather than included the proceedings of the Haskell Symposium.  Authors of accepted papers may change this decision on the basis of the reviews.
 
-Authors of accepted papers will present their papers at the Symposium.  
+Authors of accepted papers will present their papers at the Symposium, whether or not they are intended for JFP.  Authors of papers intended for JFP may also provide links to materials to be included in the program.
 
-JFP will accept the HS reviews as the first round of reviewing for JFP, provided the reviews meet the standards of the journal.  Authors of accepted papers will then be given the option to revise their papers for submission to the Journal of Functional Programming.  To assure that these appear in a timely fashion, the deadline for revision will likely be around the time of the Symposium itself.  Revised paper will be reviewed by JFP, most likely by the original reviewers.
+JFP will accept the HS reviews as the first round of reviewing for JFP, provided the reviews meet the standards of the journal.  Authors of accepted papers will then be given the option to revise their papers for submission to JFP.  To assure that these appear in a timely fashion, and to give authors the chance to include feedback from the Symposium, the deadline for revision will shorty after the Symposium itself.  Revised papers will be reviewed by JFP, most likely by the original reviewers.
 
 The final decision of the JFP editors is, of course, independent of the decision of the Haskell Symposium PC.
 
 ### For PC members
 
-Under this proposal, reviews of (accepted) papers should meet JFP's review standards.  Haskell Symposium reviewers should also be prepared to review paper revisions for JFP.
-
-### For the community
-
-A consequence of moving away from an archival proceedings is that ACM SIGPLAN may ask us to change the title of the meeting to something more like "Haskell Workshop", to line up with other language workshops held around ICFP.
-
-## Stakeholders
-
-The primary stakeholders are future authors and participants in the Haskell Symposium.    
-
-In particular, we invite feedback on whether the JFP submission process would make authors more (or less) likely to consider the Haskell Symposium as a venue for presenting work in the future.
+Under this proposal, reviews of (accepted) papers should meet JFP's review standards.  Haskell Symposium reviewers should also be prepared to review paper revisions for JFP.
--- a/proposals/0000-haskell-symposium-changes.md
+++ b/proposals/0000-haskell-symposium-changes.md
@@ -32,7 +32,7 @@ Under this proposal, authors will still submit papers to the Haskell Symposium, 
 
 Authors of accepted papers will present their papers at the Symposium, whether or not they are intended for JFP.  Authors of papers intended for JFP may also provide links to materials to be included in the program.
 
-JFP will accept the HS reviews as the first round of reviewing for JFP, provided the reviews meet the standards of the journal.  Authors of accepted papers will then be given the option to revise their papers for submission to JFP.  To assure that these appear in a timely fashion, and to give authors the chance to include feedback from the Symposium, the deadline for revision will shorty after the Symposium itself.  Revised papers will be reviewed by JFP, most likely by the original reviewers.
+JFP will accept the HS reviews as the first round of reviewing for JFP, provided the reviews meet the standards of the journal.  Authors of accepted papers will then be given the option to revise their papers for submission to JFP.  To assure that these appear in a timely fashion, and to give authors the chance to include feedback from the Symposium, the deadline for revision will be shortly after the Symposium itself.  Revised papers will be reviewed by JFP, most likely by the original reviewers.
 
 The final decision of the JFP editors is, of course, independent of the decision of the Haskell Symposium PC.
 
--- a/proposals/0000-haskell-symposium-changes.md
+++ b/proposals/0000-haskell-symposium-changes.md
@@ -1,0 +1,51 @@
+# RFC: Reinvigorating the Haskell Symposium
+
+## Abstract
+
+In recent years, submissions to the Haskell Symposium have dropped sharply, making it increasingly difficult to compile a compelling event.  The Haskell Symposium steering committee has developed two independent proposals to address this.  First, we propose to accept presentations of work-in-progress.  Second, we propose that full papers at the Haskell Symposium should be submitted to JFP for publication, rather than being published in the proceedings of the symposium.  We invite feedback from the community on how these changes would effect their decisions to submit work to the Haskell Symposium.
+
+## Motivation
+
+Over the past several years, submissions to the Haskell Symposium have dropped sharply—from 34 in 2017, just two off the all-time high, to just 15 last year.  While submissions to ICFP itself have dropped over the same window, submissions to Haskell have dropped more steeply.  This makes it increasingly difficult to compile a compelling and intellectually rigorous program.
+
+One contributing explanation for this decline is the increasing role bibliometrics plays in evaluating career academics.  Despite the quality of reviewing at the Haskell Symposium, it is listed as a C tier venue—"other ranked conference venues that meet basic standards for peer reviewed venues"—in the CoRE rankings, and is not included in the [csrankings.org](https://csrankings.org) rankings of computer science departments.
+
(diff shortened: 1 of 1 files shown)


proposals/0000-haskell-symposium-changes.md
```diff
@@ -0,0 +1,51 @@
...
+
+### For authors
+
+Under this proposal, authors will still submit papers to the Haskell Symposium, and receive reviews from the PC, on the usual schedule.    
+
+Authors of accepted papers will present their papers at the Symposium.  
+
+JFP will accept the HS reviews as the first round of reviewing for JFP, provided the reviews meet the standards of the journal.  Authors of accepted papers will then be given the option to revise their papers for submission to the Journal of Functional Programming.  To assure that these appear in a timely fashion, the deadline for revision will likely be around the time of the Symposium itself.  Revised paper will be reviewed by JFP, most likely by the original reviewers.
```
nomeata, line 35: I find that attractive.

Should the deadline be “soon after the Symposium”, so that any insights gained from both the last-minute talk preparation and especially the hallway discussions after the talk or other timely feedback from attendees can be incorporated?

ivanperez-keera: I also would like this. I'd suggest that the deadline be about 3-4 weeks after the Symposium.

Why it should be after:
- We'll gather more feedback during the conference, which can help make the paper better.
- People sometimes are busy during the weeks prior preparing their slides and getting funding and making travel arrangements, so having it at the same time can be a pain.

Why 3-4 weeks and not just 3-4 days:
- Less risk of people having to spend their ICFP week working on a paper for submission to JFP.
- We won't don't have to cut the trip short or rush on the way back. It's not unusual to take time off right after and to have some catching up to do at work after coming back from ICFP.

ivanperez-keera: > The steering committee proposes to cease publishing the proceedings of the Haskell Symposium as an archival ACM publication. [...]

> For authors
> 
> Under this proposal, authors will still submit papers to the Haskell Symposium, and receive reviews from the PC, on the usual schedule.
> 
> Authors of accepted papers will present their papers at the Symposium.

Can we really call them accepted "papers" anymore if there are no proceedings? Aren't they more like "accepted talks"?

> JFP will accept the HS reviews as the first round of reviewing for JFP, provided the reviews meet the standards of the journal. Authors of accepted papers will then be given the option to revise their papers for submission to the Journal of Functional Programming.
> 
> The final decision of the JFP editors is, of course, independent of the decision of the Haskell Symposium PC.

Could this make universities, employers and funding organizations not fund trips for submitters because there's no paper being published at the Symposium?

I'd like to propose an alternative: the same that ICFP does. Publish the symposium papers perhaps in a PACML-like journal, and invite selected papers to publish extended versions at JFP.

ivanperez-keera: Also, the benefit of doing it this way is that, by the time of the conference, there's a published paper that people can read.

What you are proposing is effectively to have post-proceedings, which is what IFL and TFP do. I don't submit there because they have post-proceedings, and there's zero guarantee of the paper being published, so as an author/employer, we acquire the cost up-front for a possible return.

Many funding organizations take paper publications as a measure of productivity. If somebody submits to the Haskell Symposium, presents the paper, and then JFP says "no" to that paper, the people who funded their trip may be livid, and they authors may have spent their whole travel budget on this event.

Another alternative would be to have the submission deadline _way earlier_, so that, by the time that the author gets to the symposium, the paper is accepted _and_ published.

simonpj:
There is a fork in the road here.

* EITHER (A): continue with the Haskell Symposium as an archival ACM publication, with full reviews etc.  Consequence: a HS paper is a publication and you can't resubmit that paper elsewhere.
* OR (B): move back to a more informal workshop format, in which there is no published proceedings, and an accepted paper or talk can certainly be submitted elsewhere for publication.

We have been doing (A) for some time, but the steering committee is signalling that (A) isn't working very well; in particular we aren't getting enough submissions.   A contributory factor could be (although it's hard to be certain) that conference ranking systems (e.g. http://www.conferenceranks.com/) don't rate the HS highly, which makes authors think about submitting elsewhere.

Hence the suggestion to move towards (B), while still providing authors with a route (but not an *obligation*, nor a *guarantee*) of an archival publication in JFP.

@ivanperez-keera points out that it is much easier to get your institution to fund a trip to ICFP if you have a publication at ICFP.   True enough.  But we also need to balance the fact that the majority of participants are not authors of accepted papers; and that the declining submission rate must be telling us something about the attractiveness of HS as an archival venue.  Just doing more of the same is possible, but personally I think we might want to be more proactive than that.

The proposal is for consultation; it's not decided!  If you have better ideas, please do offer them here.

goldfirere:
We could have our cake and eat it, too.

(Aside: I'm in strong support of having talk proposals as a separate category. I want to hear more cool ideas that are cooking in Haskell-land, even if they're not yet polished and are out-of-scope for HIW. There seems to be no controversy on this point.)

The interesting idea presented here is linking HS presentations with JFP. There is some debate, however, about the timing of the relationship. Why not have this aspect of the relationship be flexible? That is, authors who want to present their latest work at HS can submit to HS, get a review, present, and then post-submit to JFP. In addition, authors can submit to JFP directly, get their paper accepted, and then present at HS.

One question: why would a JFP author choose to present at HS rather than ICFP (who also accept JFP presentations)? This is a bit harder. We could simply open up this option and see if anyone takes us up. Perhaps someone would if they're more interested in having their travel budget include HS than ICFP. Or maybe ICFP has to be selective in inviting JFP authors who want to present -- we should ask. If they are selective, we can offer HS as a second venue for presentation. Or we could make the HS PC serve an entire year, and then JFP submitters can opt to be reviewed by members of the HS PC for their initial review, even if off-cycle. A paper accepted by this route would be presented at HS, not ICFP.

High-level piece: thanks @jgbm for thinking creatively about how to evolve the HS. I think this kind of innovation is needed.

ulysses4ever:
> There is a fork in the road here.

Not necessarily. Some conferences have been employing a mixed model for some time. [ECOOP](https://2024.ecoop.org/track/ecoop-2024-papers#Call-for-Papers), for example, has what they call Journal After with a similar strategy to what’s proposed here (there are three journals they collaborate with). If you go that route, ECOOP will still publish an extended abstract of your paper. (And you get to present it, of course.) But that’s optional: the “normal” publications are still the default option.

jmct:
> In recent years, submissions to the Haskell Symposium have dropped sharply, making it increasingly difficult to compile a compelling event. 

A compelling _two day_ event, or a compelling event, full stop?

Perhaps Haskell has reached a state of maturity where there isn't the flurry of research _about Haskell_ that there once was, and a single day event (like ML-Family, Scheme, etc) is more appropriate?

I also like the point that @ulysses4ever brought up, a mixed-mode approach could be worthwhile, _if_ we think there's enough for a good single-day event.

If the issue is that there are too few submissions for even a single day event, it might point to different problems.

Aside from that, I also feel a bit sad that bibliometrics are a factor here, but I recognize that I have the privilege of not worrying about that in the way that others must. I

ivanperez-keera:
> That is, authors who want to present their latest work at HS can submit to HS, get a review, present, and then post-submit to JFP.

@goldfirere If the paper does not appear in any proceedings, for this to count as a publication, authors will have to go to the Haskell symposium and also pay JFP which IIRC has its own fees.

adamgundry:
I wonder about the relationship between this idea and the future of the Haskell Implementors' Workshop. HIW has typically been a venue for more informal talk proposals, but I believe it has also seen a decline in the number of submissions. Moreover my impression is that ZuriHac now attracts more of the "open-source Haskell hacker" crowd than ICFP does.

Perhaps rather than having both HS and HIW co-located with ICFP over 3 days, it would make sense to have a single 1-2 day Haskell event at ICFP, and a HIW-like event alongside ZuriHac (somewhat similar to the GHC Contributors' Workshop last year)?

goldfirere:
Good point @ivanperez-keera -- we don't want to increase costs.

It might be interesting to experiment co-locating HS with ZuriHac one year. (I think the ICFP steering committee might be happy to co-locate all of ICFP with ZuriHac at some point, if ZuriHac can move to the fall for the year.) I think expanding the talk options around ZuriHac is a great idea. The only drawback is that its location is fixed, meaning that some people are always going to have a hard time getting there.

We might be able to do both here, too: have one day of HS co-located with ZuriHac, and have the other day co-located with ICFP. Authors then have a choice about which venue is better for them (including timing, location, audience).

ivanperez-keera:
My interpretation of what @adamgundry was saying, and correct me please if I'm wrong (I don't want to speak for anyone), is that the HIW aligns more with Zurihac in terms of style and topics, so maybe something like that could be co-located with Zurihac, and something like the HS could be co-located with ICFP and cover part of the spot left by the HIW.

proposals/0000-haskell-symposium-changes.md
```diff
@@ -24,28 +24,18 @@ Talk proposals will not be distributed to attendees, but authors of talk proposa
...
+Authors of accepted papers will present their papers at the Symposium, whether or not they are intended for JFP.  Authors of papers intended for JFP may also provide links to materials to be included in the program.
 
-JFP will accept the HS reviews as the first round of reviewing for JFP, provided the reviews meet the standards of the journal.  Authors of accepted papers will then be given the option to revise their papers for submission to the Journal of Functional Programming.  To assure that these appear in a timely fashion, the deadline for revision will likely be around the time of the Symposium itself.  Revised paper will be reviewed by JFP, most likely by the original reviewers.
+JFP will accept the HS reviews as the first round of reviewing for JFP, provided the reviews meet the standards of the journal.  Authors of accepted papers will then be given the option to revise their papers for submission to JFP.  To assure that these appear in a timely fashion, and to give authors the chance to include feedback from the Symposium, the deadline for revision will shorty after the Symposium itself.  Revised papers will be reviewed by JFP, most likely by the original reviewers.
```
ivanperez-keera, line 35: the deadline for revision will **be** shortly after...

jgbm:
I think the arguments for continuing to have an archival proceedings of the Haskell Symposium are compelling (for now).

The updated proposal allows 3 kinds of submissions:

1. Talk proposals
2. Full papers, for publication in HS
3. Full papers, for publication in JFP

Everything accepted will be presented at HS.  Accepted full papers in category 2 will be included in the proceedings of the Haskell Symposium.  Accepted full papers in category 3 will continue to JFP, as proposed before.

@goldfirere raises a challenging point: given that JFP articles can appear in ICFP, why would an author *who was convinced that their paper was JFP-ready* prefer to present at the Haskell Symposium?

One answer I can imagine is timing: for work that just misses the ICFP deadline, it might be preferable to present the work sooner, while still having credit for publication in a higher-tier venue.

There have been several suggestions about relocating either part of the HS, the HIW, combining with related events, and the like.  I think these are interesting suggestions, but I think there would be value in seeing how the number of submissions changes with these proposals before proposing further (and more dramatic) changes.

@adamgundry points out that this could lead to HIW talks being sent to HS instead.  I hope this is not the case.  Looking at last year's HIW talks, many of them focus specifics of implementation that might not be suited to HS.  In the longer term, I can see it making sense to reduce HS to one day, and have HIW as the second day of Haskell-related content at ICFP.

simonpj:
@jgbm proposals sounds sensible to me.

sgraf812:
> @goldfirere raises a challenging point: given that JFP articles can appear in ICFP, why would an author who was convinced that their paper was JFP-ready prefer to present at the Haskell Symposium?

From what I gather, it's not that the JFP article "appears in ICFP" in the sense of an ICFP proceedings publication; it's just a talk slot.
I had a brief look at the titles of all JFP talks from ICFP 2020 to ICFP 2023; the only talk that was seemed specific to Haskell was [Modal FRP for all: Functional reactive programming without space leaks in Haskell](https://www.cambridge.org/core/journals/journal-of-functional-programming/article/modal-frp-for-all-functional-reactive-programming-without-space-leaks-in-haskell/9BE20E8D61E9B74811CF3CF97B5D10C7). I think the risk of cannibalisation is reasonably low.

proposals/0000-haskell-symposium-changes.md
GrahamHutton: I agree with Ivan that authors may not be keen on presenting a paper that isn't guaranteed to lead to a publication.  There's a danger that not getting a guaranteed publication makes the submission rate worse, rather than better.  One possible solution is to use a two-phase reviewing process prior to the paper being accepted for the Haskell Symposium and JFP publication, which is how special issues of the journal tend to be run, e.g. http://tinyurl.com/mr37duxh.  But the downside is that this process takes more time, although it could be compressed for the purposes of the symposium rather than a special issue.

jmct:
If @jgbm feels that the Steering Committee has gotten what it wants out of this RFC, we propose archiving the discussion. We're going to create a process for archiving RFCs that will make it easy to review the discussion, while closing the PR itself.

However, if you feel that there's more to discuss, we can delay archiving it and try to get more eyes on the RFC.

jgbm:
@jmct I think archiving this discussions makes good sense.

PR #64, opened 2024-01-19
Botan Cryptography Community Project Proposal Extension

ldillinger:
This is a draft proposal for the extension of funding for the [Botan Cryptography Community Project](https://github.com/apotheca/tech-proposals/blob/main/proposals/052-cryptography-leg-1.md).

The TWG may make comments and feedback as necessary, or may vote immediately if they wish.

### DIFF
--- a/proposals/053-cryptography-extension.md
+++ b/proposals/053-cryptography-extension.md
@@ -1,0 +1,81 @@
+# Botan Cryptography Community Project Extension
+
+> NOTE: This document references an earlier proposal that is ongoing and close to meeting its goals. Language in this document may reflect a future perspective that assumes those goals have been met.
+
+# Abstract
+
+The Botan Cryptography Community Project is reaching the end of its funding on Jan 31, 2024. This proposal is for the extension of funding to and addition of new goals for the [Botan Cryptography Community Project](https://github.com/haskellfoundation/tech-proposals/pull/57), which is a previously funded proposal that is ongoing, and nearing completion of its initial goals. These new goals are for the continued refinement of the produced libraries, completion of the X509 C++ work, and many other small features and minutia. 
+
+# Background
+
+The [Botan Cryptography Community Project](https://github.com/apotheca/botan) is a project that was previously funded by the Haskell Foundation for the purpose of developing a suite of Haskell cryptography libraries and tooling suitable a wide range of uses including data integrity, privacy, security, and networking.
+
+This project has maintained a [devlog](https://discourse.haskell.org/t/botan-bindings-devlog/6855) and produced several monthly reports [(1)](https://discourse.haskell.org/t/botan-cryptography-monthly-status-report-0/8280) [(2)](https://discourse.haskell.org/t/botan-cryptography-monthly-status-report-1/8497)
+
+The project has produced 3 libraries: `botan-bindings`, `botan-low`, and `botan`.
+
+This project also has made a substantial amount of progress towards one of its stretch goals of extending the Botan C FFI to include full X509 support. We were initially optimistic that this stretch goal might be completed by the end of initial funding, but the C++ work required more overhead than expected, and efforts towards it were paused to avoid endangering any required deliverables.
+
+# Technical Content
+
+Despite (being on pace for) fulfilling its goals, there is still more work to do, as the scope of the original proposal was deliberately restricted for the purposes of focus and achievability. Now that the original goals have been met, some of the previously out-of-scope goals may now be reconsidered.
+
+## Continued improvements to `botan`
+
+- `botan-bindings` and `botan-low` are effectively frozen, but small changes may still occur
+- `botan` is likely to see significant churn as we continue to improve and respond to feedback
+
+As `botan` is a very "fresh" library, it is expected to see some amount of churn in its API as it is developed and refined further. The initial goals of functionality are nominally complete, but there is still a great amount of minutia to be performed that is concerned with improving the safety and user experience. Expected improvements include limited lifespan objects with guaranteed cleanup, auto-zeroing of sensitive memory, and implementing instances from libraries such as `random`.
+
+## Botan C++ development
+
+- `Botan C++` is being actively developed, so project will need continued maintenance to achieve parity
+- Several valuable features in `Botan C++` are not exposed by the C FFI, and require C / C++ development in order to expose them
+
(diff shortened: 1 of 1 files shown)


ldillinger:
This proposal has been updated with a retrospective, the addition / refinement of several goals, and improved clarity / detail of expected deliverables.

[Rendered](https://github.com/haskellfoundation/tech-proposals/blob/5c712d9f9e428d9a25d19effce02668490aada4b/proposals/053-cryptography-extension.md)

ldillinger:
Minor update regarding mitigating factors.

[Rendered](https://github.com/haskellfoundation/tech-proposals/blob/a7fcaa7248950078642fe18a24b060797e0f554d/proposals/053-cryptography-extension.md)

proposals/053-cryptography-extension.md
```diff
@@ -0,0 +1,124 @@
...
+
+- Stream ciphers
+- Complete X509 infrastructure implementation
+- Complete TLS server implementation
+
+One of the previous stretch goals was to develop the C bindings for X509 certificate architecture, and to [contribute that code back](https://github.com/randombit/botan/issues/3627) to `Botan C++`. Substantial progress was made, but further work is required in order to complete this. 
+
+Another feature missing from the Botan C FFI is Stream Ciphers, and like the X509 bindings, it will require effort to extend the Botan C++ library in order to expose it to Haskell.
+
+A third feature missing from the Botan C FFI is TLS, as Botan provides a full TLS server implementation, which will also require effort to expose to the FFI.
```
Bodigrim, line 69: I understand the rationale, but essentially the proposal asks HF to sponsor a development of a third-party C library, which we have no control of / do not gain any stake in / are at mercy of its leadership. This is a high risk investment, extremely difficult to manage. Could we rather focus on lower-hanging fruits and wait (potentially indefinitely) until someone else does Botan C FFI?

proposals/053-cryptography-extension.md
```diff
@@ -0,0 +1,124 @@
...
+
+Due to the insufficiency of the ADT approach towards managing algorithms, cryptographic typeclasses and data families have been brought back into scope as a deliverable after being declared out-of-scope in the original proposal.
+
+While a first attempt has already been made to typeclassify the various cryptographic primitives, this interface will require extensive refinement and feedback over the coming months, and is likely to experience significant change before settling down.
+
+It is expected that these typeclasses will result in improved per-algorithm type safety, and better ergonomics through more consistent handling of cryptographic primitives such as keys, nonces, and ciphertexts.
+
+## Improvements to installation / CI / Unit tests
+
+One of the largest warts in using these libraries is the installation of the Botan C++ library. Package distributions are available for some users and operating systems via `pkg-config`, but other users currently require manual installation.
```
Bodigrim, line 81: I think at the moment the only OS which can install Botan via package manager is macOS (`brew install botan`), while everyone else has to build from sources.

proposals/053-cryptography-extension.md
```diff
@@ -0,0 +1,124 @@
...
+    - To develop an abstract cryptographic typeclass / data family hierarchy exhibiting per-algorithm safety
+    - To improve Botan installation / CI ease of use
+
+The following are intermediate deliverables:
+
+    - Frequent activity update on devlog
+    - Weekly major update that includes github push
+    - Monthly report that summarizes what specific items have been or will be worked on, and what challenges have arisen, if any.
+
+The following are stretch goals / optional deliverables to be looked into only after required deliverables are met:
```
Bodigrim, line 102: How many stretch goals could be complete, assuming an optimistic scenario? I feel like 5 goals are beyond possible even if everything goes very smooth, so I'd suggest to whittle down the list to a more realistic shape.

Bodigrim:
My overall preference would be to focus on a specific, user-facing, tangible deliverable. For example, can we name X such that "upon the completion of this proposal package X could switch from `cryptonite` to `botan`"? Given that this is a second leg, I'd expect us to deliver vertical integration instead of horizontal expansion.

mpickering:
I think when I read this proposal it isn't clear to me what the community benefit of such a project is. It sounds like the initial funding has been used to complete the original goals, so now I presume that the library can be used by the original stakeholders. Job done?

I agree with @Bodigrim that any more horizontal expansion seems a bit premature. Without users already using the existing library, it begs the question, at which point is the community going to benefit from investing a very significant amount of money into this project?

Perhaps we shouldn't engage in what-aboutery, as you have taken the effort to write this proposal and get this work funded, but I can think of many other projects which would benefit from $20k USD which would have a more immediate community impact.

hasufell:
I don't see that Botan is in a state where it doesn't need active development anymore. This is a crypto library and not some PoC that should be abandoned half way in.

Wrt the concrete goals I agree with @Bodigrim that working towards making it a drop-in replacement with good documentation/guides and comparisons is reasonable. That, however, may also require working on new features.

ldillinger:
Appreciations for your patience, I have taken a week off for rest, and have not yet had a chance to respond yet.

This feedback *is* being paid attention to, and I will be updating this proposal with more concrete details about what I intend to accomplish in the near-term future / within the scope of this proposal.

gbaz:
> I think when I read this proposal it isn't clear to me what the community benefit of such a project is. It sounds like the initial funding has been used to complete the original goals, so now I presume that the library can be used by the original stakeholders. 

I think the original proposal had a much more thorough motivation and this should just be seen as add-on work which was already motivated as possible stretch goals or long term goals in the original proposal.

The overall reason this work is important and impactful is we have crypto libraries sitting at the root of a _lot_ of production haskell code and more generally dependency trees for anything involving networking, the web, apis, etc. The existing libraries are idiosyncratic, not audited, and not well maintained. Having solid reliable well maintained crypto libraries is a huge benefit to the ecosystem, and having _some working libraries at all_ is necessary for essentially every industrial use case.

ldillinger:
This proposal was updated significantly in tandem with a pending [First Milestone blog post](https://github.com/haskell-cryptography/haskell-cryptography.github.io/pull/4) of the same subject, but for a different audience / the general community.

ldillinger:
Minor updates to the proposal: reducing libsodium interface to long-term goal, clarified CI improvements / `build-type: configure` goals

Tritlo:
I like the proposal, and I do think it is worthwhile. However, you speak of "continued maintenance" for parity with the Botan C++ library, but still you only seek funding for 3 months of work. Can we assume that a further proposal for "continued maintenance" past these 3 months for later? Or we could possibly have some sort of "maintenance" clause or similar in this one.

Just thoughts! I'm still in favour for accepting this one for the other parts of the work.

